PR #4, opened 2020-08-24
Separate badges from tags RFC and added rfc/

allisonsuarez:
Added rfc to propose creating a new node type for neo4j which would represent Badges that is separate from the Tag node.


rfc/004-feature-template.md
```diff
@@ -0,0 +1,54 @@
+- Feature Name: badges_node
+- Start Date: 2020-08-24
+- RFC PR: [amundsen-io/rfcs#4](https://github.com/amundsen-io/rfcs/pull/4)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000)
+
+# Separate badges from tags
+
+## Summary
+
+Create a new node type for neo4j which would represent Badges that is separate from the Tag node.
```
feng-tao, line 10: Amundsen metadata has different backends (neo4j, atlas, gremlin is coming). I think `creating a separate entity for badge` would be more suitable for our case.

rfc/004-feature-template.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Motivation
+
+In order to best use badges we want to add sentiment and category fields to the Badge nodes. This makes the Tag and Badge components different, as Badge requires more information now. Separating the two now will likely lead to less tech debt in the future when other features are implemented that impact the way we ingest and use badges. Additionally, this change will help clarify the difference in use and purpose between tags and badges throughtout Amundsen. This feature will also guarantee badges are not ingested without having sentiment and category.
```
feng-tao, line 14: It would be good to : 
- provide context on what badge is, 
- cover what use case it tries to address, 
- discusses why we did previously to combine two into tag 
- discusses what new use case it tries to address and why we think separate them is better than adding two new attributes

rfc/004-feature-template.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+## Guide-level Explanation (aka Product Details)
+
+> Explain the proposal as if it was already included in Amundsen and you were teaching it to an Amundsen user. That generally means:
```
feng-tao, line 19: does section need to update ? If not, we may just delete.

rfc/004-feature-template.md
```diff
@@ -0,0 +1,54 @@
...
+> If applicable, provide deprecation warnings, or migration guidance.
+> For implementation-oriented RFCs, this section should focus on how maintainers should think about the change, and give examples of its concrete impact. For policy RFCs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms.
+- In the future this feature will hopefully replace all use of old Tag badges.
+
+## Reference-level Explanation (aka Technical Details)
+
+The implementation will require:
+- Add methods to the metadata service API to add, remove and, get badges as currently the service uses the methods for tags.
+- Update badge ingestion docs to reflect new API call and parameters.
+- Make updates to search, databuilder and frontend service to add functionality to use Badge instead of Tag.
```
feng-tao, line 33: I think we could expand the details here with a little bit more context : how search is handling badge and what is the new proposed , same for fe, databuilder, etc?

We could have the implementation prs going in the mean time while keep updating these section while ready. WDYT?

rfc/004-feature-template.md
```diff
@@ -0,0 +1,54 @@
...
+- Takes more effort and time to implement than the alternative of keeping them together.
+- Duplicate code for some functionality between BadgeAPI and TagAPI.
+
+## Alternatives
+
+- Keep the current functionality and simply add two optional parameters fo sentiment and category to Tag.
+
+## Unresolved questions
+
+- Do we want to constraint the input space for sentiment and category to be an enum to reduce room for error rather than being a string?
```
feng-tao, line 50: another question would be do we plan to add more attributes for badge later (e.g besides type, category)? Does the current proposed model support that?

rfc/004-feature-template.md
feng-tao: also should we cover column level badge discussion here?

rfc/004-feature-template.md
Golodhros: I think it should contain what you asked for in the previous comment!

rfc/004-feature-template.md
allisonsuarez: I think that's a good idea since I don't yet have enough context about how search and databuilder, I will add more detail about FE now.

feng-tao:
> > will take a look. Also the name of the doc should be 001-badge.md
> 
> I believe the instructions said that the name of the document should match the PR number

could you point me to the instruction? I think we need to update that part of the instruction as we need to start the RFC from 001, Let me know if you think differently. cc @Golodhros @ttannis

feng-tao:
@Golodhros looking at the readme ```Copy the appropriate template file to rfcs/000-<rfcName>.md. Don't assign an RFC number yet, as this is going to be the PR number, and we'll rename the file accordingly if the RFC is accepted.```, any reason we need to rename after it is accepted instead of creating from the beginning?

rfc/004-feature-template.md
feng-tao: yeah, that's why I think once you touch all the places(FE, search, metadata, databuilder) where badge is used, you could update this doc correspondingly. In the end, I think RFC should be used to document the context of why, where, and how we propose the change.

rfc/004-feature-template.md
```diff
@@ -7,23 +7,18 @@
...
-> Explaining the feature largely in terms of examples.
-> Explaining how Amundsen users should think about the feature, and how it should impact the way they use Amundsen. It should explain the impact as concretely as possible.
-> If applicable, provide deprecation warnings, or migration guidance.
-> For implementation-oriented RFCs, this section should focus on how maintainers should think about the change, and give examples of its concrete impact. For policy RFCs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms.
-- In the future this feature will hopefully replace all use of old Tag badges.
+A badge is a predetermined indicator which can be added by a table owner in order to offer informartion about the satus of a table resource (alpha, beta GA, deprecated), weather they recommend using a certain resource, as well as to indicate if a column in a table is a primary key or partition column. Using these to boost results in searches and to filter resources when searching will help Amundsen users find the resources they need easier, as well as help them discern which resources are most trustworthy.
```
Golodhros, line 21:
```suggestion
A badge is a predetermined indicator which can be added by a table owner in order to offer information about the status of a table resource (alpha, beta GA, deprecated), whether they recommend using a certain resource, or to indicate if a column in a table is a primary key or partition column. 

Using Tags to boost results in searches and to filter resources when searching will help Amundsen users find the resources they need easier, as well as help them discern which resources are most trustworthy.
```

Golodhros:
> > > will take a look. Also the name of the doc should be 001-badge.md
> > 
> > 
> > I believe the instructions said that the name of the document should match the PR number
> 
> could you point me to the instruction? I think we need to update that part of the instruction as we need to start the RFC from 001, Let me know if you think differently. cc @Golodhros @ttannis

Yes, the idea is that people don't need to look at what number of RFC is, or to handle collisions, as the numbers of the PRs will be unique. The results would mean that we have skipped numbers, but they will always be on the right order.

Golodhros:
> @Golodhros looking at the readme `Copy the appropriate template file to rfcs/000-<rfcName>.md. Don't assign an RFC number yet, as this is going to be the PR number, and we'll rename the file accordingly if the RFC is accepted.`, any reason we need to rename after it is accepted instead of creating from the beginning?

Now that yo mention it, I can't see any reason why we should wait. I still think is a good idea to use the PR number though, as mentioned on the previous comment.

feng-tao:
> > @Golodhros looking at the readme `Copy the appropriate template file to rfcs/000-<rfcName>.md. Don't assign an RFC number yet, as this is going to be the PR number, and we'll rename the file accordingly if the RFC is accepted.`, any reason we need to rename after it is accepted instead of creating from the beginning?
> 
> Now that yo mention it, I can't see any reason why we should wait. I still think is a good idea to use the PR number though, as mentioned on the previous comment.

I am not sure if I like the skip number for rfc though as we could always look at which pr for a given RFC based on git blame ? Other popular projects I know didn't use skip number : https://cwiki.apache.org/confluence/display/AIRFLOW/Airflow+Improvements+Proposals or https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Improvement+Proposals

sorry that I miss this part in the initial review, but I am curious how others think of this. @ttannis or @amundsen-io/amundsen-committers

rfc/004-feature-template.md
```diff
@@ -0,0 +1,51 @@
...
+## Drawbacks
+
+- Takes more effort and time to implement than the alternative of keeping them together.
+- Duplicate code for some functionality between BadgeAPI and TagAPI.
+
+## Alternatives
+
+- Keep the current functionality and simply add two optional parameters fo sentiment and category to Tag.
+
+## Unresolved questions
```
allisonsuarez, line 44: What is the impact of changing the badges field int ables/dashboards to contain
`badge = {
    'badge_name': fields.String,
    'category': fields.String,
    'badge_type': fields.String,
}`
rather than having it be like tags (tag_type, name)?

allisonsuarez: @ttannis where in the frontend did you say that metadata is gotten? I am looking at https://github.com/amundsen-io/amundsenfrontendlibrary/blob/ccfd2d6b82957fef347e956b243e4048c191fc0d/amundsen_application/api/metadata/v0.py and see that there is no mention of badges, only tags.

allisonsuarez: For more context making a request like ` curl -X GET curl -X GET http://0.0.0.0:5002/"hive://gold.events/event_client_ux"/`
with what I have currently implemented on local ([PR](https://github.com/amundsen-io/amundsenmetadatalibrary/pull/176)) the response would include both fields badges and new_badges:
```   "badges": [
        {
            "tag_type": "badge",
            "tag_name": "allison"
        }
    ],```

```
    "new_badges": [
        {
            "category": "table_status",
            "badge_name": "test_dash",
            "badge_type": "neutral"
        }
    ],
```

@feng-tao suggested that we only have badges which has all badges shown according to the new_badge model. I would like to know how this change would impact other services in Amunden?

Comment (ttannis) [reply rfc/004-feature-template.md:44]: I see. The work sounds like it would be as follows on FE:
1. Return all badges as `badges` according to the new badge model.
2. For backwards compatibility in the community, some folks could still be using Tags of `tag_type: "badge"` and we would have to account for their case. As a result we would have to include temporary logic in all of the `marshall_<some-resource>` methods in https://github.com/amundsen-io/amundsenfrontendlibrary/blob/master/amundsen_application/api/utils/metadata_utils.py. That helper method would have to look for objects of the old shape, and replace them with objects of the new shape.
3. Then the React application can write UI logic assuming the new shape and the new shape only. This would also include updating our `Tag` type interface and creating a new one for the badge object as well so that we can leverage TypeScript as usual to catch any errors when operating on the objects. 
4. Once we confirm we can officially deprecate `tag_type: "badge"` , then the code introduced in Step 2 can be deleted.

There could be some edge cases I've missed on regards to perhaps filtering support and the search service, but I'm confident we'll be able to deal with them and will advise along the way.

Comment (ttannis) [reply rfc/004-feature-template.md:44]: However this does mean that until we deprecate Tags of `tag_type: "badge"`, apis that need badge information will also have to return tags so that FE can pull out and convert all the `tag_type: "badge"`

Comment (allisonsuarez) [reply rfc/004-feature-template.md:44]: We could put those in a legacy_badge field or just add logic that converts all of them to the new badge functionality with default values for category and badge_type like 'table_status' and 'neutral' respectively.

Comment (feng-tao) [reply rfc/004-feature-template.md:44]: I wonder whether we should just do a quick check on the community and see if anyone is using badge and need to have this support. Also we could always cut a break change release between all three services and document it in a doc (previously we have done sth similar when we update the model in https://github.com/amundsen-io/amundsen/blob/master/docs/faq.md#how-to-migrate-from-amundsen-1x---2x)

The hack that does the conversion in FE  is sth I would like to see if we could avoid... WDYT @allisonsuarez @ttannis ?

Comment (allisonsuarez) [reply rfc/004-feature-template.md:44]: Yeah I would really like to know what the current use cases for the community are so we can update accordingly and clean up anything that is not being used. I am nut sure what you mean by "cut" a break change @feng-tao

Comment (feng-tao) [reply rfc/004-feature-template.md:44]: oh, I meant cut a release version for all three services that have the breaking change.

Comment (ttannis) [reply rfc/004-feature-template.md:44]: ^ we could. But I also thought that something that would be frustrating for others is creating a new major version for individual features. There are several other items in our code base that would also benefit from being included in a planned major release, where we dedicate time to properly fixing backwards incompatible changes, deprecating configurations, etc. 

While these helper methods could seem hacky, I've seen it used elsewhere at a previous job. What made it work well and feel less hacky is that these things were well organized into separate "migration methods", and the official release process for major releases included a task to address removing the "migration methods" for what features were scheduled to be deprecated in that release. 

Overall though, I'm fine to go with whatever option the majority of maintainers support. I acknowledge that there's a chance we could get off easy this time around if the community isn't currently heavily using Tags of `tag_type: "badge"`.

File: rfc/004-feature-template.md [hunk 8]
```diff
@@ -0,0 +1,51 @@
...
+- RFC PR: [amundsen-io/rfcs#4](https://github.com/amundsen-io/rfcs/pull/4)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000)
+
+# Separate badges from tags
+
+## Summary
+
+There is a need to create a new entity for Badges that is separate from Tag in the metadata service. Badges in Amundsen are components which in the past have been used to indicate different bits of metadata like explaining if a resource was a table view or if a user was no longer active in amundsen (alumni). Tags were also simple bits of information consisting of a name that were used to filter resources when searching, so it made sense to combine them at the time. 
+
+Moving forward we want to use badges to indicate the satus of a table resource (alpha, beta GA, deprecated), weather we recommend using a certain resource, as well as using them to indicate if a column in a table is a primary key or partition column.
```
PaschalisDim, line 12: Small typo here (satus-> status)

feng-tao:
@allisonsuarez given the feature is almost there, let's finish the rfc doc when you have time and merge it?

bogo96:
After the column badge PR of databuilder, i can't see the ```hive``` table badge. 
The `table_metadata` model still use only tags. I think `Badge` model should be added to `table_metadata` model.
Is there any plan to fix it?

allisonsuarez:
> After the column badge PR of databuilder, i can't see the `hive` table badge.
> The `table_metadata` model still use only tags. I think `Badge` model should be added to `table_metadata` model.
> Is there any plan to fix it?

Interesting, I don’t think I made any change to table level tags that would cause this. The reason you would not be able to see it in the UI is because we no longer surface tags with the type badge. Was this badge being extracted on databuilder and being put on the table as a tag of type badge? Can you not see it on neo4j or is it just not visible on front end?

rfc/004-feature-template.md
```diff
@@ -0,0 +1,48 @@
+- Feature Name: badges_node
```
feng-tao, line 1: hey @allisonsuarez , could you change the rfc name to be 001-support-badge-as-entity.md and change the folder to be rfcs?

feng-tao (commented):
lgtm with one nit

rfc/004-feature-template.md
feng-tao: the doc is still using the template name.

Golodhros:
Let's make this into the **Final Comment Period (FCP)** to _merge,_ so there will be seven days for anyone to comment in this PR before we merge it and make it official.

bogo96:
> > After the column badge PR of databuilder, i can't see the `hive` table badge.
> > The `table_metadata` model still use only tags. I think `Badge` model should be added to `table_metadata` model.
> > Is there any plan to fix it?
> 
> Interesting, I don’t think I made any change to table level tags that would cause this. The reason you would not be able to see it in the UI is because we no longer surface tags with the type badge. Was this badge being extracted on databuilder and being put on the table as a tag of type badge? Can you not see it on neo4j or is it just not visible on front end?

- https://github.com/amundsen-io/amundsendatabuilder/blob/af4b51265e11d1ee9e8651cebeb0d00592eec2d5/databuilder/extractor/neo4j_search_data_extractor.py#L36
   : According to query, it seems `Table` has `Badge` model

-  https://github.com/amundsen-io/amundsendatabuilder/blob/master/databuilder/models/table_metadata.py#L242
   : But still, `TableMetadata` has only tag parameter and there's no parameter to make relation with `Badge` model.

- https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/metadata_service/api/tag.py#L60
   : Also, there are no api to getting table badge.

allisonsuarez:
> > > After the column badge PR of databuilder, i can't see the `hive` table badge.
> > > The `table_metadata` model still use only tags. I think `Badge` model should be added to `table_metadata` model.
> > > Is there any plan to fix it?
> > 
> > 
> > Interesting, I don’t think I made any change to table level tags that would cause this. The reason you would not be able to see it in the UI is because we no longer surface tags with the type badge. Was this badge being extracted on databuilder and being put on the table as a tag of type badge? Can you not see it on neo4j or is it just not visible on front end?
> 
> i can see that table has badge type tag on neo4j, but it is not visible on frontend.
> i think we can't use badge type tag because of below codes.
> 
> * https://github.com/amundsen-io/amundsendatabuilder/blob/af4b51265e11d1ee9e8651cebeb0d00592eec2d5/databuilder/extractor/neo4j_search_data_extractor.py#L36
>   : According to query, it seems `Table` has `Badge` model
> * https://github.com/amundsen-io/amundsendatabuilder/blob/master/databuilder/models/table_metadata.py#L242
>   : But still, `TableMetadata` has only tag parameter and there's no parameter to make relation with `Badge` model.
> * https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/metadata_service/api/tag.py#L60
>   : Also, there are no api to getting table badge.

That's because we [announced to the community](https://amundsenworkspace.slack.com/archives/CGFBVT23V/p1598985853019000) that we would be deprecating this type of badge in favor of the new badge model that was implemented. What you can do is delete your old badges using the delete tag API method on the amundsenmetadatalibrary, and then use this tutorial to add the new badges to those tables: https://github.com/amundsen-io/amundsen/blob/79bbfbd0899daed5a0d203fb523ebd6b73842c92/docs/tutorials/badges.md

bogo96:
> > > > After the column badge PR of databuilder, i can't see the `hive` table badge.
> > > > The `table_metadata` model still use only tags. I think `Badge` model should be added to `table_metadata` model.
> > > > Is there any plan to fix it?
> > > 
> > > 
> > > Interesting, I don’t think I made any change to table level tags that would cause this. The reason you would not be able to see it in the UI is because we no longer surface tags with the type badge. Was this badge being extracted on databuilder and being put on the table as a tag of type badge? Can you not see it on neo4j or is it just not visible on front end?
> > 
> > 
> > i can see that table has badge type tag on neo4j, but it is not visible on frontend.
> > i think we can't use badge type tag because of below codes.
> > 
> > * https://github.com/amundsen-io/amundsendatabuilder/blob/af4b51265e11d1ee9e8651cebeb0d00592eec2d5/databuilder/extractor/neo4j_search_data_extractor.py#L36
> >   : According to query, it seems `Table` has `Badge` model
> > * https://github.com/amundsen-io/amundsendatabuilder/blob/master/databuilder/models/table_metadata.py#L242
> >   : But still, `TableMetadata` has only tag parameter and there's no parameter to make relation with `Badge` model.
> > * https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/metadata_service/api/tag.py#L60
> >   : Also, there are no api to getting table badge.
> 
> That's because we [announced to the community](https://amundsenworkspace.slack.com/archives/CGFBVT23V/p1598985853019000) that we would be deprecating this type of badge in favor of the new badge model that was implemented. What you can do is delete your old badges using the delete tag API method on the amundsenmetadatalibrary, and then use this tutorial to add the new badges to those tables: https://github.com/amundsen-io/amundsen/blob/79bbfbd0899daed5a0d203fb523ebd6b73842c92/docs/tutorials/badges.md

Oh, i'll check tutorial docs. thanks!

feng-tao (approved):

feng-tao:
@allisonsuarez let's merge it ?

Golodhros:
> @allisonsuarez let's merge it ?

@allisonsuarez do you want to do the honors?

allisonsuarez:
> > @allisonsuarez let's merge it ?
> 
> @allisonsuarez do you want to do the honors?

I got out of bed for this lol

feng-tao:
lol

PR #8, opened 2020-10-21
ML project discovery

dwarszawski:


000-ml-project-discovery.md
```diff
@@ -0,0 +1,81 @@
+- Feature Name: ml_project_discovery
+- Start Date: 2020-10-21
+- RFC PR: [amundsen-io/rfcs#000-ml-project-discovery.md](https://github.com/amundsen-io/rfcs/pull/0000) (after opening the RFC PR, update this with a link to it and update the file name)
```
Golodhros, line 3:
```suggestion
- RFC PR: [amundsen-io/rfcs#008-ml-project-discovery.md](https://github.com/amundsen-io/rfcs/pull/8)
```

000-ml-project-discovery.md
```diff
@@ -0,0 +1,81 @@
...
+
+The basic use cases can be described as follow:
+
+*   search for ml projects by project name, project description, project owner
+*   search for ml models by model name, model description, model owner, model stage
+*   search for model deployments by deployment name, deployment description, deployment sample request/response (prediction)
+*   view the data sources used to build the model 
+*   view the metadata of the model, basic evaluation metrics e.g. accuracy, precision, recall, f1 etc.
+
+## UI/UX-level Explanation
```
feng-tao, line 32: @dwarszawski do you or does ING have any designer which could help to drive certain UX mock for the model? that will be great

ttannis: Let's also collaborate on it either way, without dedicated design support this part might move slower but I can help make a best effort to align what is needed with existing design concepts.

dwarszawski: Apologize for late response. @feng-tao @ttannis I couldn't agree more. We should follow up on the UX part. Maybe next community meeting is a good time to start discussion?

ttannis: For the most part it'd be valuable to first have an understanding of all the generic kinds of metadata that apply to ML features, what is read-only vs editable etc. It'd be great to see the community discuss that first, and then consider the UI after that's settled.

000-ml-project-discovery.md
```diff
@@ -0,0 +1,81 @@
...
+The concept of machine learning project refers to complex process of moving your research from local exploration environment towards production grade deployments.
+Essentially, that consists of multiple steps including feature engineering, tracking experiments, model training, batch scoring, model versioning, model staging, release and deployment.
+Many organizations suffers from lack of visiblity over the ML assets delivered by the different teams of data scientists.
+Such model catalog would improve the collaboration within the organization through single interface to discover available ML resources and related metadata.  
+
+The basic use cases can be described as follow:
+
+*   search for ml projects by project name, project description, project owner
+*   search for ml models by model name, model description, model owner, model stage
+*   search for model deployments by deployment name, deployment description, deployment sample request/response (prediction)
```
mgorsk1, line 28: @dwarszawski does it mean introducing 3 additional entities on top of tables and dashboards ?

000-ml-project-discovery.md
dwarszawski: We are considering ml discovery and lineage as the essential parts of the ml project cycle. Different kind of metadata can be recognized at various stages of the project. From feature engineering (set of columns with description, statistics, normalization methods) through model training (number of iterations, batch size, optimizer, loss function etc) to model deployment (model owner, use case description,  stage of the model, performance reports).

000-ml-project-discovery.md
dwarszawski: it's continuously evolving, I think it's better to describe it in terms of the metadata available during different stages of ml project lifecycle.

joshthoward:
@dwarszawski After reading through this, I would not conflate ML projects with ML models. The current ontology includes data assets (i.e. tables, dashboards) which might be the outcome of a generic project. I believe it would make more sense to add a model asset and related entities through this RFC and separately consider about how projects should be incorporated.

mgorsk1:
agree with @joshthoward , probably ML model is in the center of everything and it would be the easiest entity type to get the ball rolling. wdyt @dwarszawski

joshthoward:
To be clear... I think projects are a great idea, but I think that they would extend beyond ML.

dwarszawski:
> @dwarszawski After reading through this, I would not conflate ML projects with ML models. The current ontology includes data assets (i.e. tables, dashboards) which might be the outcome of a generic project. I believe it would make more sense to add a model asset and related entities through this RFC and separately consider about how projects should be incorporated.

dwarszawski:
@joshthoward imho "ML model" term is not a precise enough. However, it may be considered as the top level abstraction over any outcome of ML training process  - could be just a serialized model but also entire pipeline, scoring results (table), or just a code depending on what is expected on the downstream site.

dwarszawski:
so models can fall into two types: 
- ml pipeline - set of transformations on raw data to get the scores
- ml model - serialized model for real time predictions

PR #11, opened 2020-11-04
Configurable Elasticsearch Queries

mgorsk1:


feng-tao:
@mgorsk1 qq: did ING use ES for search as well? I thought both metadata and search is based on atlas? I heard about this request from Square before, but the concern would be :1. only for power users; 2. how do it fit into multiple search backend?

Do you also have a few custom search ES query examples in your mind that could help improve the search quality?

mgorsk1:
We are in the middle of PoC aimed to replace Solr search with ES. ES might be used as search backend instead of Solr but it still requires using Atlas API for search and nothing beats flexibility of ES Search API so we want to check if using it directly will bring positive change to our search experience.

So far I've been thinking on positive/negative boosting based on tags/badges - like:
```
{
  "query": {
    "boosting": {
      "positive": {
        "term": {
          "tag": "official"
        }
      },
      "negative": {
        "term": {
          "tag": "deprecated"
        }
      },
      "negative_boost": 0.5
    }
  }
}
```

Second approach (instead of config option) would be to refactor ES proxy to fetch queries from text files. Heavy users could then just overwrite them before running search proxy.

mgorsk1:
@feng-tao @Golodhros reviving this topic. Do you think moving queries to proxy config would be sufficiently non-invasive approach here ?

feng-tao:
I will take a look and get back to this RFC next week.

feng-tao:
one more question: do you think we still need Atlas search proxy in the future? If we are going to make ES as the only proxy for search, we could support more fancy search request .

mgorsk1:
That's a good question. In our case I think we are happy with ES proxy to the point of not looking back. The only downside for this route one could notice is that Atlas Search proxy doesn't require any additional work (over deploying Apache Atlas) while ES requires you to:
1. Deploy Atlas with ES which is not official & well documented route from Atlas perspective (Solr is the default option).But it's easily doable - we've upstreamed all the changes for janusgraph queries to work ok.
2. Schedule search synchronization - Atlas <> ES search synchronization works without any issues, it's just additional component.
3. If you already deployed Atlas with Solr then you need to have 2 search engines to maintain in parallel, which is not ideal

I'd go with documenting our new approach in Amundsen docs and suggest using ES over Atlas for search. cc @verdan

PR #12, opened 2020-11-10
Proposes the use of Python's Dataclasses

verdan:

Python 3.7 introduced [dataclasses (PEP557)](https://docs.python.org/3/library/dataclasses.html). Dataclasses can be a convenient way to generate classes whose primary goal is to contain values. 

We are relying on `attrs` for (de-)serializing data for each request within Amundsen. `attrs` itself is a big package with so many different features available that we don't even need. 

By using Python's native dataclasses we can have one less dependency with almost the same features `attrs` provide that we intend to use.

### DIFF
--- a/rfcs/000-python-dataclasses.md
+++ b/rfcs/000-python-dataclasses.md
@@ -1,0 +1,88 @@
+- Feature Name: python-dataclasses
+- Start Date: 2020-11-10
+- RFC PR: [amundsen-io/rfcs#0000](https://github.com/amundsen-io/rfcs/pull/0000) (after opening the RFC PR, update this with a link to it and update the file name)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
+
+# Leverage Python's (native) Dataclasses instead of attrs (a third party package)
+
+## Summary
+
+- Python 3.7 introduced [dataclasses (PEP557)](https://docs.python.org/3/library/dataclasses.html) .Dataclasses can be a convenient way to generate classes whose primary goal is to contain values. 
+- The design of dataclasses is based on the pre-existing attr.s library, and are basically a slimmed-down version of attrs. 
+
+- In fact `Hynek Schlawack`, the very same author of attrs, helped with the writing of PEP557.
+
+## Motivation
+
+We are relying on `attrs` for (de-)serializing data for each request within Amundsen. 
+`attrs` itself is a big package with so many different features available that we don't even need. 
+By using Python's native dataclasses we can have one less dependency with almost the same features `attrs` provide that we intend to use.
+
+## Guide-level Explanation (aka Product Details)
+
+-  The details about the dataclasses can be found here: https://docs.python.org/3/library/dataclasses.html
+-  Example of a dataclass:
+```python
+from dataclasses import dataclass, asdict
+
+@dataclass
+class InventoryItem:
+    """Class for keeping track of an item in inventory."""
+    name: str
+    unit_price: float
+    quantity_on_hand: int = 0
+
+    def total_cost(self) -> float:
+        return self.unit_price * self.quantity_on_hand
+
+inventory = InventoryItem(name='alpha', unit_price=20.0, quantity_on_hand=5)
+
+# OR directly from a dict
+data = {"name": 'alpha', "unit_price": 20.0, "quantity_on_hand": 5}
+inventory_alt = InventoryItem(**data)
+
+# Can access fields directly by dot notations
+assert inventory_alt == "alpha"
+
+# and then convert it back to dict 
+asdict(inventory_alt)
+```
+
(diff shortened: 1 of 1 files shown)


feng-tao:
currently Amundsen supports python36 and above. Will it force to drop python36 support?

verdan:
> currently Amundsen supports python36 and above. Will it force to drop python36 support?

@feng-tao Yes, it will. I've made it explicit in the technical details and also in the drawbacks. 

I believe we should aim to support as many python versions as possible, but MUST support at least the last 3 versions. The current Python version is 3.9, so I think it is okay to drop the support of Python3.6.

feng-tao:
Given python36 is still widely used these days while python39 is just introduced, I will be hesitant to just drop python36. I take a quick look at airflow and superset which both still supports py3.6.

feng-tao:
@verdan actually it is backported to py36 (https://github.com/ericvsmith/dataclasses). I think it is fine from the python version pov.

rfcs/000-python-dataclasses.md
```diff
@@ -0,0 +1,88 @@
...
+complex, instead look for a simpler and cleaner solution. 
+
+
+## Prior art
+N/A
+
+## Unresolved questions
+
+- Should we aim for having as less dependencies as possible? That includes the third party packages. 
+- Do we think of a case where we'll be doing heavy data validations, and if those can not be done with dataclasses?
```
feng-tao, line 84: I think it is ok.

verdan:
> @verdan actually it is backported to py36 (https://github.com/ericvsmith/dataclasses). I think it is fine from the python version pov.

oh, this is nice. I did not know that before. I am still curious however why companies are still using Python3.6, as Python3.7 introduced a number of nice feature that the community was waiting for. but yes, the major share is still Python3.6 as you mentioned.

dorianj:
Strongly in support of this -- cleaner syntax and typing, and the python3.6 backport works well. Any reason not to push this through and start using?

verdan:
@feng-tao do we agree to move this to the `Active` state, and start implementing this if no further comments?

feng-tao:
sgtm

jornh:
To quote https://github.com/amundsen-io/rfcs/pull/12#issuecomment-725277808

> I believe we should aim to support as many python versions as possible, but MUST support at least the last 3 versions. The current Python version is 3.9.

Not really related to this RFC itself, but since the topic is up. We're currently only officially supporting 3.6 and 3.7. Looks like the Databuilder package works for folks with 3.8, but tests need some love + maybe addition to CI. Similarly the 3 services need some work just to be able to build with 3.8.

Time to kick off that work? I'd be happy to do some of it. I guess it doesn't need a RFC as it's basically just a maintenance chore, not introducing any new feature or breaking anything.

rfcs/000-python-dataclasses.md
```diff
@@ -0,0 +1,88 @@
...
+## UI/UX-level Explanation
+
+Nothing changes on the UI/UX.
+
+## Reference-level Explanation (aka Technical Details)
+
+- The primary change will happen in amundsencommon repositories where we are defining the models. I believe no other major change 
+will be required in any other repository. 
+
+- DROPS SUPPORT OF PYTHON 3.6
```
jornh, line 60:
```suggestion
- Python 3.6 can still be supported as long as it's needed through  backported to py36 ([ericvsmith/dataclasses](https://github.com/ericvsmith/dataclasses)). 
```

rfcs/000-python-dataclasses.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Drawbacks
+
+- One of the main thing to consider is the Python 3.6 support. If we go for the dataclasses, we will not be supporting 
+Python 3.6 anymore. The minimum version requirement for Python for Amundsen will become 3.7.
+
```
jornh, line 66:
```suggestion
```

verdan:
> To quote [#12 (comment)](https://github.com/amundsen-io/rfcs/pull/12#issuecomment-725277808)
> 
> > I believe we should aim to support as many python versions as possible, but MUST support at least the last 3 versions. The current Python version is 3.9.
> 
> Not really related to this RFC itself, but since the topic is up. We're currently only officially supporting 3.6 and 3.7. Looks like the Databuilder package works for folks with 3.8, but tests need some love + maybe addition to CI. Similarly the 3 services need some work just to be able to build with 3.8.
> 
> Time to kick off that work? I'd be happy to do some of it. I guess it doesn't need a RFC as it's basically just a maintenance chore, not introducing any new feature or breaking anything.

@jornh that would be awesome if you could please help out with making sure that all the services work with the latest 3 versions of Python. 

In the meantime, I am taking care of the dataclasses.

dorianj:
Note that dataclasses were landed in the front-end repo, with the back-compat library for 3.6, and is working smoothly.

dorianj (approved):

PR #14, opened 2020-11-10
Format all Python code across Amundsen codebases with Black

Spacerat:
Following https://github.com/amundsen-io/amundsendatabuilder/pull/405 - @feng-tao asked me to create an RFC, so here it is!

### DIFF
--- a/rfcs/003-use-black.md
+++ b/rfcs/003-use-black.md
@@ -1,6 +1,6 @@
 - Proposal Name: Use the Black formatter to format all Python code
 - Start Date: 2020-11-10
-- RFC PR: [amundsen-io/rfcs#13](https://github.com/amundsen-io/rfcs/pull/0000) (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: [amundsen-io/rfcs#13](https://github.com/amundsen-io/rfcs/pull/14)
 - Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
 
 # <RFC title>
--- a/rfcs/003-use-black.md
+++ b/rfcs/003-use-black.md
@@ -1,0 +1,96 @@
+- Proposal Name: Use the Black formatter to format all Python code
+- Start Date: 2020-11-10
+- RFC PR: [amundsen-io/rfcs#13](https://github.com/amundsen-io/rfcs/pull/0000) (after opening the RFC PR, update this with a link to it and update the file name)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
+
+# <RFC title>
+
+## Summary
+
+Format all Python code across Amundsen codebases with [Black](https://black.readthedocs.io/en/stable/).
+
+## Motivation
+
+### Motivation for code formatting in general
+
+As an open source contributor - and in general, a contributor to any codebase - having to think and care about code formatting adds an extra hurdle to contribution.
+While writing code, you have to ask yourself "is this conforming to the style of the code around it?". We use flake8, but its standards are very lax and it can't fix
+anything for you, so when you think you're done coding, you have to go through a tedious loop of "lint -> fix -> lint -> fix" until everything is good. If a PR needs
+any changes, the cycle begins again.
+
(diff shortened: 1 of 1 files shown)


rfcs/003-use-black.md
```diff
@@ -0,0 +1,96 @@
+- Proposal Name: Use the Black formatter to format all Python code
+- Start Date: 2020-11-10
+- RFC PR: [amundsen-io/rfcs#13](https://github.com/amundsen-io/rfcs/pull/0000) (after opening the RFC PR, update this with a link to it and update the file name)
```
Golodhros, line 3:
```suggestion
- RFC PR: [amundsen-io/rfcs#13](https://github.com/amundsen-io/rfcs/pull/14) (after opening the RFC PR, update this with a link to it and update the file name)
```

Golodhros:
I support this 100%! Thanks for the RFC and the PR!

rfcs/003-use-black.md
```diff
@@ -0,0 +1,96 @@
...
+Google uses YAPF.
+
+## Unresolved questions
+
+- Should we include a pre-commit hook?
+- Will we resolve the `git blame` issue? And how?
+- If we choose Black, should we go with a line-length of 120?
+    - I chose that in my PR because flake8 had been configured that way
+    - Since the Databuilder repo seems prone to some particularly long lines, I think 120 probably makes sense; but I'm ambivalent.
+- Which repos will this apply to?
```
dorianj, line 89: in additiojn to databuilder, I i think we should apply to amundsenfrontendservice, amundsenmetadataservice, amundsensearchservice and common -- as mentioned, I'm happy to do the implementation, but I think we should standardize across all of the repos

feng-tao: yeah, if we decide to enforce linter, we should apply to all the other repos as well.

Spacerat: The more the merrier! I've only been working closely with Databuilder, which is the reason that's where my focus has been

verdan:
I'm not a big fan of autoformatters personally because of two main reasons:
- Autoformatted code is not always readable code, which makes it hard to maintain. They remove the previous formatting (the natural format by a developer). They don't understand how each specific case will be most readable, as they rely only on one style all the time for each scenario. This should always be the developer's concern.

Consistency is definitely important within a project, a repo, and most importantly within a class/function. However, consistency does not always mean readability. We can leverage the strict linting rules to enforce a `style guide` across all repositories of Amundsen, that won't allow the bad styled code in the master branch in the first place. 

PEP8 has a section that I 100% agree with regarding the readability https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

- (_this is a personal factor involving continuous learning_): I prefer to write a well-formatted code instead of having a tool fixing my bad code for me. Autoformatters fix your code without you knowing why it did what it did. So you are not learning the best practices and community standards at all. 

That being said, if we, as a community decide to go for an autoformatter, I'd be inclined towards yapf, as it does not add so many extra white spacing and extra lines in the code blocks. This is more like enforcing consistency with no major format changes.

Golodhros:
> I'm not a big fan of autoformatters personally because of two main reasons:
> 
> * Autoformatted code is not always readable code, which makes it hard to maintain. They remove the previous formatting (the natural format by a developer). They don't understand how each specific case will be most readable, as they rely only on one style all the time for each scenario. This should always be the developer's concern.

I am not super familiar with Black, but if its code is not readable, that is a problem! I guess when you get used to it, as the format is always the same, it will become more familiar and thus, readable.

> * (_this is a personal factor involving continuous learning_): I prefer to write a well-formatted code instead of having a tool fixing my bad code for me. Autoformatters fix your code without you knowing why it did what it did. So you are not learning the best practices and community standards at all.

I used to be in agreement with you, but lately I have changed my mind. Using prettier in the frontend has saved me a lot of time, specially as it reformats on save, leaving me more time to work on the real problems and speeding up my development. I think you will always see what the formatter did when you review your own PRs, and you can always check why that formatting was applied (at least that's the case with JS and eslint).
 
> That being said, if we, as a community decide to go for an autoformatter, I'd be inclined towards yapf, as it does not add so many extra white spacing and extra lines in the code blocks. This is more like enforcing consistency with no major format changes.

Maybe this calls for a maintainer's vote then? I'm happy with whatever you backenders decide.

feng-tao:
let's do some evaluations first unless we feel comfortable to know different linter pros/cons. I personally haven't looked into the difference between those projects and what rules they try to enforce on top of pep8.

PR #15, opened 2020-11-23
Frontend Library Split RFC

Golodhros:
Proposal for a split of the frontend library


rfcs/00x-frontend-library-split.md
```diff
@@ -0,0 +1,88 @@
...
+
+Not doing this would further leave us behind the latests improvements on the field, will continue having a complex setup and onboarding, making it hard to contribute to and hard to extend our platform.
+
+## Prior art
+
+We have talked several times about this, never tried it. It has been a common move in all internal projects at Lyft though.
+
+## Unresolved questions
+
+We'll need to create a detailed step-by-step plan for this refactor.
```
feng-tao, line 77: I think this needs to be settle in a concrete plan before we discuss the RFC.

rfcs/00x-frontend-library-split.md
```diff
@@ -0,0 +1,88 @@
...
+## Future possibilities
+
+For the backend:
+
+- Easily open the backend to other services and applications
+
+For the frontend:
+
+- Update it to use a static site generator like Gatsby
+- Update it to become a Next.js application, incorporating best practices and best performance
```
feng-tao, line 88: why next.js?

rfcs/00x-frontend-library-split.md
```diff
@@ -0,0 +1,88 @@
+- Feature Name: frontend-library-split
+- Start Date: 2020-11-23
+- RFC PR: [amundsen-io/rfcs#150(https://github.com/amundsen-io/rfcs/pull/15) (after opening the RFC PR, update this with a link to it and update the file name)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
+
+# Split of the frontendlibrary repository in two
+
+## Summary
+
+At this moment, our frontend repository contains a Flask application that acts as an API server and also serves a frontend application based in React.js. This means that there are two applications (Flask and Frontend), with two languages (Python and JavaScript), two configuration settings, two dependency management systems (pip and npm), two sets of owners (frontend and backend developers), and two main "responsibilities" (API server and User Interface application) baked into the same repository. This is a problem, as it makes the system complex, couples both systems and stifles the evolution of both applications.
```
feng-tao, line 10: Currently Amundsen includes three services, the FE is just one of them. If we split, any reason we need the API server instead of calling metadata/search directly.

rfcs/00x-frontend-library-split.md
```diff
@@ -0,0 +1,88 @@
...
+
+We propose to split our current amundsenfrontend repository into a backend-only API server and a frontend-only application.
```
feng-tao, line 12: same question as above.

rfcs/00x-frontend-library-split.md
```diff
@@ -0,0 +1,88 @@
...
+
+We consider this the best approach as it lines up with general best practices, decouples our system and adds flexibility while reducing complexity.
+
+One alternative would be to do this same work while keeping both in the same repository. It won't bring all the benefits, but it will be a bit better than the current status.
+
+Not doing this would further leave us behind the latests improvements on the field, will continue having a complex setup and onboarding, making it hard to contribute to and hard to extend our platform.
+
+## Prior art
+
+We have talked several times about this, never tried it. It has been a common move in all internal projects at Lyft though.
```
feng-tao, line 73: But Apache superset and Redash (both popular open source bI viz projects) choose this arch for FE.

feng-tao (commented):
I think the benefit of split is unclear(e.g Apache superset, redash both chooses flask + React) while the backward incompatible migration overhead is sth concerned.

rfcs/00x-frontend-library-split.md
Golodhros: Agreed!

rfcs/00x-frontend-library-split.md
Golodhros: Just an example, but is it one of the most popular platforms right now. They help by setting a lot of defaults out of the box, making it easier to setup a new application. It also provides a good amount of interesting features from the start.

rfcs/00x-frontend-library-split.md
Golodhros: That's a great question, maybe we could remove this step and call metadata/search directly.
I see it as a good "proxy" server for the functionality on those services, and I guess it would be great if we could make it extensible through plugins, so we could implement things like the emailing feature we have at Lyft.

rfcs/00x-frontend-library-split.md
Golodhros: I am not familiar with those project requirements or specifications, as well as their context and the expertise on the frontend that they had on the time of their creation. I can only talk about Amundsen's issues and context.

rfcs/00x-frontend-library-split.md
jcarey03: I see the potential value of the API server as a proxy. Are there other use cases we know of that would benefit from a proxy?

jcarey03 (commented):
Thanks @Golodhros for kicking this off! I'm interested to learn more about the reduced operational costs and use cases that will benefit from this.

rfcs/00x-frontend-library-split.md
```diff
@@ -0,0 +1,88 @@
...
+Why we shouldn't do this?
+
+- Load of work involved
+- Migration costs for current users
+
+## Alternatives
+
+We consider this the best approach as it lines up with general best practices, decouples our system and adds flexibility while reducing complexity.
+
+One alternative would be to do this same work while keeping both in the same repository. It won't bring all the benefits, but it will be a bit better than the current status.
```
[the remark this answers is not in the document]
Golodhros: Maybe, but I think it would be more complex.

rfcs/00x-frontend-library-split.md
```diff
@@ -0,0 +1,88 @@
...
+- Allow for multiple frontends
+- Streamlined development
+
+### Evolution of applications
+
+Having two applications in different repositories will reduce the coupling, allowing us to evolve both apps independently, implementing the newest best practices on both. For example, this would make it easier to create a GraphQL based interface for our API, as well as adopting a full stack structure based in Next.js or a static site based in Gatsby in the frontend.
+
+### Allow for multiple frontends
+
+Extracting the API server into its own applications would allow us to create different frontends for the Amundsen application. Think about creating a mobile application, a desktop app or forking the current UI application to customize it deeply to the user's specifications.
```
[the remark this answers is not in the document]
Golodhros: It would be indeed useful to know how many users have forked completely their own frontend library without working on top of the standard. Our current configuration options help a bit with the customization, but they can only go so far.

dorianj:
At a high level, I'd rather see the project move to a monorepo structure (possibly slowly, as needed). Much has been spilled on the internet about this, but at a high level I think it would:

- Increase velocity — right now one change might require 2-4 different PRs across different repos, making the overhead of making and reviewing changes much higher.
- Decrease tooling cost — right now CI/CD/test efforts are replicated between 4-5 different repos. For a mostly-volunteer army, this is a lot of work to maintain and implement. Very easy for lint settings to get out of line, etc.
- Decrease barrier to entry for OSS contributors, make modifications easier, etc. Right now even downloading and forking is not trivial, since you have to deal with git submodules, many people don't know how to do it because it is not common.
- Make code sharing easier. I know we have the shared repo, but practically writing code in such a way that you need to open 2 PRs and link them means the barrier to actually using the shared repo is very high. The amundsencommon repo is not used very much, I believe because of the hassle involved — engineers naturally will use the path of least resistance.
- Customization and configuration would be easier. It would be easier to share config files between different projects. If people want to do patch-based modifications to their own install, it would be easier with a single repo rather than maintaining per-repo patches. Forks are easier too since you can fork the entire thing rather than maintain many forks.

Monorepo CI/CD tooling is quite good these days, and I think things like slow build or test times are fairly easily avoidable with the right tools. The historical advantages of deployability granted by single repos are diminished since deploys are generally done with Docker anyhow. Codeowners can make the review distribution somewhat automatic. I think most newer OSS projects use monorepos these days for all of their code. 

**node.js**

I've done a fair share of both node and python web services that serve JS front-ends. I haven't noticed a massive productivity jump in having the same FE and BE language both be in JS: most full stack engineers are familiar enough with both that it's not a huge gap. Given that we already have python services elsewhere, I think staying with python makes the most sense, especially if we can share backend data models. Particularly, If we had some way to do data modeling that was shared by both the FE and the back-end, I might change my opinion slightly, but I think that would be a pretty massive endeavor and possibly hard to do incrementally, and generally has less great tooling than python backend data modeling.

**Calling metadata store directly**

I'm generally in favor of the [Backend-for frontend pattern](https://samnewman.io/patterns/architectural/bff/) — I think eventually to retain high performance on the UI, you end up needing to either make monstrously generic APIs that are hard to maintain, or make specific calls for the UI at hand. I think the latter approach generally results in more-maintainable code and overall faster velocity for product engineering teams. Having APIs that are shared between a bunch of different clients makes it really hard to change existing APIs, slowing down development for all the UIs. But I do suspect this is all theoretical for now — do we know of anyone who has done custom or plans to do custom UIs?

So that would suggest having a layer similar to the current one in place. But it's possible that's better achieved by the metadata service having some per-UI custom routes, rather than the front-end service being an entirely different endpoint. There definitely are a lot of places where the front-end service simply takes a request, parses it, and sends it over to the metadata service. If that were in the same service, there would be a lot fewer extraneous tests and layers to add when adding functionality. Basically how that would look is the metadataservice exposes some core primitives on both a code level and as RPC APIs (or RESTful or whatever). Then the front-end APIs would use those code interfaces, rather than having to bounce back to HTTP. I suspect that this would also make authorization security easier to apply.

rfcs/00x-frontend-library-split.md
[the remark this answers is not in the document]
markgrover: There's only one company in the community who has created their own - to match their own look and feel. That's Cimpress (video and notes [here](https://docs.google.com/document/d/1MT3qd_YjFiA17en94IEZcWaEgjdcaLhhZb-mX1UyWF8/edit#)). I don't think it's a super common use-case, but I can double check if you'd like.

There are others that have added some minor to moderate changes to UI that haven't been merged to master. On minor changes, examples are REA group and iRobot adding new structured metadata columns (for things like primary key, or link to example Mode reports respectively), and Square adding a human in the loop to approve/reject suggested PII tags ([PDF doc in slack](https://amundsenworkspace.slack.com/files/UJ35GMPUZ/FSR2GE141/square_demo___amundsen.pdf), [blog](https://developer.squareup.com/blog/using-amundsen-to-support-user-privacy-via-metadata-collection-at-square/)). They have forked the UI because they think our focus is on data discovery, while their use-cases are more in privacy engineering/data governance. I think that's reasonable.

So, all in all, I think it's super rare. My potentially controversial opinion: I won't over-index on the forking use-case. Thoughts?

Golodhros:
> At a high level, I'd rather see the project move to a monorepo structure (possibly slowly, as needed). 
+100 to this! I just created a JS monorepo lately and it is way more effective and would make working in Amundsen more efficient. 
However, I don't know how to manage a mixed python/JS monorepo, I guess a lot of stuff needs to be done manually vs the JS only projects using Lerna, right? 
 
> **node.js**
I wasn't really proposing to change the API server to be node based, although this change would allow to do it easily if we wanted to go that way.

> So that would suggest having a layer similar to the current one in place. But it's possible that's better achieved by the metadata service having some per-UI custom routes, rather than the front-end service being an entirely different endpoint. 
That sounds good to me. The migration would be easier too.

dorianj:
> +100 to this! I just created a JS monorepo lately and it is way more effective and would make working in Amundsen more efficient.
> 
> However, I don't know how to manage a mixed python/JS monorepo, I guess a lot of stuff needs to be done manually vs the JS only projects using Lerna, right?

I haven't done this exact combination in a few years... but if there's some bubbling support for it, I'd be happy to propose an RFC and do the necessary research.

> I wasn't really proposing to change the API server to be node based, although this change would allow to do it easily if we wanted to go that way.

ohh, sorry, i misread the next.js mention. Makes sense.

rfcs/00x-frontend-library-split.md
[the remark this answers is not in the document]
Golodhros: Thanks for the explanation @markgrover!

Golodhros:
Hi team,
To wrap up this RFC, here is a summary:
- We agree that the migration path would be the most important challenge when doing something like this
- Most of us think that keeping an API proxy layer would be useful for now
- All of this would be a lot easier if we were in a monorepo setup

Because of that, I will close this RFC and instead create another one to suggest moving Amundsen into a monorepo configuration, then we would be able to do the FE split a lot easier.

PR #17, opened 2020-12-09
Migrate to Official Python Atlas Client

verdan:

Atlas community recently released an official python client to interact with Apache Atlas. For now, we are relying on our custom client, which does not have all the endpoints implemented yet.

### DIFF
--- a/rfcs/000-migration-official-atlas-client.md
+++ b/rfcs/000-migration-official-atlas-client.md
@@ -1,6 +1,6 @@
 - Feature Name: migrate_to_official_atlas_client
 - Start Date: 2020-11-09
-- RFC PR: [amundsen-io/rfcs#0000](https://github.com/amundsen-io/rfcs/pull/0000) (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: [amundsen-io/rfcs#17](https://github.com/amundsen-io/rfcs/pull/17) (after opening the RFC PR, update this with a link to it and update the file name)
 - Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
 
 # Migrate to Official Python Atlas Client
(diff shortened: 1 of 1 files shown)


feng-tao:
@verdan lgtm! I think it is mostly up to you to decide:
1. % of feature parity between your client and atlas official one.
2. if we want to contribute back atlas python client, how often they do python release considering it is from an Apache project.

verdan:
We do have a few helper functions we added in our version of client, which will be missing in the official client. But those are mostly for the Amundsen, and we will move them to amundsen repos. 

> if we want to contribute back atlas python client, how often they do python release considering it is from an Apache project.

This is my main concern, as the Atlas community is sometimes slow in releasing new versions. However the good thing is that the first version is shipped with all the existing endpoints, so this is already better then the one we are using. 😄

pPanda-beta:
I found several bugs in the official client. Not even in a mood to create "Given-when-then-but" fashion of bug reports for those. 

We need to ensure the client is stable enough before migrating.

verdan:
@mgorsk1 @pPanda-beta We need to decide how we want to proceed with this. I already have the working implementation of the Atlas Proxy using the official client. All the code to Atlas is already upstreamed. As Palash mentioned it's a bit tricky with not having any test cases at all. On the other hand, the investment to rewrite our test cases in Amundsen atlas proxy is already a huge investment of time. 
So we need to decide if we really want to do this without the proper test cases in the official atlas package.

mgorsk1:
I'd expect that this rfc always assumed tests need to be refactored for new client. Couldn't be avoided I guess and I think we would be fine just having tests on our side.

verdan:
True, and also the missing endpoints in `pyatlasclient` also needs to be implemented which is again a separate effort (because of the complex structure of pyatlaslient) if we stick with the current implementation.

@pPanda-beta do you agree on using the official client then, and moving our implementation effort away from the `pyatlasclient`?

mgorsk1:
Another question is - will rewriting/refactoring tests make it possible to use them in the official client too ?

verdan:
> Another question is - will rewriting/refactoring tests make it possible to use them in the official client too?

Nope. We'll be writing unit tests of our functions assuming the methods of the official client are returning everything correctly by mocking them.

mgorsk1:
reviving this one as we don't want it to become blocker for upcoming features (i think tag/badge refactor is blocked by this)

cc @verdan @pPanda-beta

dorianj:
Can this be merged at this point, or does it need to go through a vote? It seems like https://github.com/amundsen-io/amundsenmetadatalibrary/pull/247 is pretty close to getting over the line

mgorsk1 (approved):
lgtm

PR #18, opened 2020-12-10
WIP: Merge Amundsen metadata and search repositories in to a single `amundsenbackend` repository

verdan:

The idea is to deprecate the split of metadata and search repositories and merge them into a single `amundsenbackend` repository.

### DIFF
--- a/rfcs/000-merge-metadata-search.md
+++ b/rfcs/000-merge-metadata-search.md
@@ -1,0 +1,54 @@
+- Feature Name: merge_metadata_search_repos
+- Start Date: 2020-12-10
+- RFC PR: [amundsen-io/rfcs#0000](https://github.com/amundsen-io/rfcs/pull/0000) (after opening the RFC PR, update this with a link to it and update the file name)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
+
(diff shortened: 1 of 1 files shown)


Golodhros:
I think this RFC proposal reinforces the idea that we are having difficulties with the current split of our repos and the sluggish workflow we need to follow in order to fix even the smallest thing.

In this sense, I agree with @Dorianj that the best first step would be to move into a monorepo, as he details on [this comment](https://github.com/amundsen-io/rfcs/pull/15#issuecomment-733289622).

feng-tao:
I will take a look at the comment but a few things for mono repos:
1. how do we make sure backward compatibility ? 
2. Users have lots of custom endpoint API in metadata/search (e.g  Lyft)? Are we able to grow metadata to a standalone API server still?
This is a key as not everyone uses Amundsen purely through UI but leverage service level integration.
3. How we do release? Having separate repos certainly do have challenges with integration but it allows something like what we did (gremlin repo, rds repo) instead of make the mono repo a kitchen sink.

verdan:
this is a WIP :thought_balloon:. I will complete the transition path, along with technical details, code splits, and even code structure on the modular level soon. This change will indeed need changes in the CI/CD pipelines and docker files but eventually would not affect much on the frontend-backend integration.

verdan:
Closed in favour of https://github.com/amundsen-io/rfcs/pull/30

PR #19, opened 2020-12-15
Support RDB Models

crazy-2020:
Support models for relational DB use

### DIFF
--- a/rfcs/000-support-rdb-models.md
+++ b/rfcs/000-support-rdb-models.md
@@ -1,6 +1,6 @@
 - Feature Name: Support Relational Database Models
 - Start Date: 2020-12-14
-- RFC PR: [amundsen-io/rfcs#0000](https://github.com/amundsen-io/rfcs/pull/0000) (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: [amundsen-io/rfcs#0019](https://github.com/amundsen-io/rfcs/pull/0019)
 - Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
 
 # Support Relational Database Models
--- a/rfcs/000-support-rdb-models.md
+++ b/rfcs/000-support-rdb-models.md
@@ -1,0 +1,122 @@
+- Feature Name: Support Relational Database Models
+- Start Date: 2020-12-14
+- RFC PR: [amundsen-io/rfcs#0000](https://github.com/amundsen-io/rfcs/pull/0000) (after opening the RFC PR, update this with a link to it and update the file name)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
+
+# Support Relational Database Models
+
+## Summary
+
+In order to use relational DBs(RDB), like MySQL, as the metadata store in Amundsen, we have to support Object-relational Mapping(ORM) first. 
+This RFC is mainly about the models used for relational DBs.
+
+## Motivation
+
+Relational DBs, like MySQL, are well deployed and widely used in most workplaces. Using the well-established DB infrastructure would save much cost(e.g., new DB deployment, backup, monitoring, proxy) 
+for potential Amundsen users. Hence, it would be very helpful if Amundsen can support relational DBs as the metadata store. 
+The expected result is users are able to call the relational DB models to ingest/retrieve metadata in MySQL.
+
+## Guide-level Explanation (aka Product Details)
+
+Instead of duplicating extractors to work with RDB models directly, RDB models are designed to work with current graph db models 
+and build each model instance as record with the extracted metadata. These models will be used in databuilder(e.g., ETL and publishing job), and metadataservice.
+
+
+## UI/UX-level Explanation
+
+N/A
+
+## Reference-level Explanation (aka Technical Details)
+
+Node and some relation labels in graph DB are reflected as the entities and associated entities in RDB models. 
+Attributes in graph node and relation is kept as the fields in RDB models. Node `key` is converted to `rk`(record key) to avoid the keyword use in some relational DBs. 
+Following are the Model(**Primary Key**, field\_1, field\_2, <ins>Foreign Key</ins>) designed according to the existing graph db nodes and relations.
+
+> Database, Cluster, Schema
+
+- Database(**rk**, name, published\_tag, publisher\_last\_updated\_epoch_ms)
+- Cluster(**rk**, name, <ins>database\_rk</ins>, published\_tag, publisher\_last\_updated\_epoch_ms)
+- Schema(**rk**, name, <ins>cluster\_rk</ins>, published\_tag, publisher\_last\_updated\_epoch_ms)
+- SchemaDescription(**rk**, description\_source, description, <ins>schema\_rk</ins>, published\_tag, publisher\_last\_updated\_epoch_ms)
+
+> Table
+
+- Table(**rk**, name, is\_view, <ins>schema\_rk</ins>, published\_tag, publisher\_last\_updated\_epoch_ms)
+- TableDescription(**rk**, description\_source, description, <ins>table\_rk</ins>, published\_tag, publisher\_last\_updated\_epoch_ms)
+- TableProgrammaticDescription(**rk**, description\_source, description, <ins>table\_rk</ins>, published\_tag, publisher\_last\_updated\_epoch_ms)
+- TableTimestamp(**rk**, last\_updated\_timestamp, timestamp, name, <ins>table\_rk</ins>, published\_tag, publisher\_last\_updated\_epoch_ms)
+- TableTag(**<ins>table\_rk</ins>**, **<ins>tag\_rk</ins>**, published\_tag, publisher\_last\_updated\_epoch_ms)
+- TableBadge(**<ins>table\_rk</ins>**, **<ins>badge\_rk</ins>**, published\_tag, publisher\_last\_updated\_epoch_ms)
+- TableOwner(**<ins>table\_rk</ins>**, **<ins>user\_rk</ins>**, published\_tag, publisher\_last\_updated\_epoch_ms)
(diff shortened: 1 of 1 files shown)


dorianj:
I agree strongly on the motivation: running Amundsen on a relational database would definitely make it more accessible and easier to operate for many teams. 

I'd take that motivation a bit further and propose that ideally, all deploys would use RDBMS as the core persistence layer. Then the graph database would be an optional component to enhance the performance of certain types of queries. That would distribute workloads to the most efficient available processor, and would mean that teams could layer on complexity of deployment as is appropriate for the scale of their deploy. Very few teams have experience with graph databases already, and in this world, only teams who were really pushing Amundsen would need to learn how to operate graph databases.

It would also be positive in terms of velocity for the Amundsen project. For new features, 90% of the work could be done in terms of RDBMS models, with types, familiar SQL syntax, etc. Then 10% of the work would be accelerating queries needing a graph database. The current code specific to graph databases would thus be reduced a lot.

But I believe this RFC is more of a fit within the current model, i.e. adding RDBMS as another source of truth layer alongside neo4j/Neptune/Atlas. At first glance, I'm opposed to adding this, for 3 reasons:
- Makes feature development slower: every new data type would now need to be 100% implemented in both neo4j/Neptune as well as RDBMS.
- I think it will ossify the current architecture wherein databuilder must write directly to the persistence stores (rather than writing through an intermediate services), making it harder to make changes in the future
- Finally, I think it will make it harder to operate Amundsen, particularly if we expect to continue making schema changes in Amundsen.

Elaborating on the last bullet, because databuilder is a library, not a service, there are a lot schema changes that we could make (e.g. simply adding a new `NOT NULL` field) that would require a coordinated databuilder upgrade. There may be different databuilder installs at a company, meaning this requires multiple components to be upgraded simultaneously (otherwise they break). If metadata service was instead the sole writer to the persistence store, this could be ameliorated as a single service would need to be upgraded atomically.

I don't want to let "good" be the enemy of "great" here by pushing back on this, in favor of something that is admittedly a lot more work and a fundamental change to the way Amundsen is set up. But I also am concerned that this may paint us in a corner that's hard to get out of. Curious for your thoughts!

crazy-2020:
Thanks for the comment. Actually, the way to support relational db support is similar to other backend stores, like Neptune. It needs new data loader, publisher, serializer for ingestion in databuilder and proxy code to handle requests from frontend in metadatservice. But on top of that, in [amundsenrds](https://github.com/amundsen-io/amundsenrds), new RDB models will be added to support ORM in databuilder and metadataservice. In order to make the implemntation easier, RDB model instances will come from the attributes in current model instead of any conversion of node/relation, i.e. The RDB model instance, namely, the RDB record, will be serialized, loaded, and then published, which is same as the current ingestion logic in databuilder. 

Certainly, the tough thing is to make any data definition change compatible in two types of DBs. Considering the schema change in relational DB, especially for metadata, we will manage RDB models through its version migration. It's not very flexible, though that's a common feature we must have to support if we use RDB as persistence layer. 

I am grateful for any other proposals to support RDBMS. This RFC was designed to make RDB as an another option for backend store. TBH, I haven't thought about distributing most workloads to RDBMS, in which case, RDBMS seems to be the sole metadata backend store. I am not sure if that means an overhaul change in Amundsen since all users are relying on Graph db now.

feng-tao:
@dorianj I won't go into the design choice on why we choose graph dbs initially. There are many features with graph db that way more appeal/performant than rds. And I think one of important factor to make Amundsen tick in OSS is the pluggable backend compared to other OSS tool(e.g linkedin wherehows in mysql).  This RFC is to add yet another persistent layer using rds(e.g mysql) for those who can't have a graph solution or use a managed graph db solution(e.g aws neptune, neo4j cloud service etc). 

For your comments:

`Makes feature development slower: every new data type would now need to be 100% implemented in both neo4j/Neptune as well as RDBMS.` -- That's not true. Different proxies could support the feature differently as long as they agree with proxy. E.g team behind lyft drove the neo4j, ING drove atlas, and square for neptune.`
`I think it will ossify the current architecture wherein databuilder must write directly to the persistence stores (rather than writing through an intermediate services), making it harder to make changes in the future` -- I think starting with RDS like the neptune case in a separate repo helps to make the dev velocity faster. But once the model is stable, we could move it to metadata service and make databuilder calls metadata put api.  
Finally, I think it will make it harder to operate Amundsen, particularly if we expect to continue making schema changes in Amundsen. -- Like previous point, I think we could move it to metadata in the model.

dorianj:
> I am grateful for any other proposals to support RDBMS. This RFC was designed to make RDB as an another option for backend store. TBH, I haven't thought about distributing most workloads to RDBMS, in which case, RDBMS seems to be the sole metadata backend store. I am not sure if that means an overhaul change in Amundsen since all users are relying on Graph db now.

It would definitely be an overhaul-type change, since Amundsen installs today to rely solely on graph DB. I hadn't thought about the transition path towards that sort of architecture in detail, but thinking more, I actually believe that deploying this changeset would be the first step towards deprecating neo4j as the source of truth.

If we did want to go in that direction, after this was shipped and had some battle-testing, we could introduce tooling for existing users to migrate from neo4j to RDBMS, and ultimately deprecate neo4j as a primary datastore and instead use it as an optional performance enhancer.

> Considering the schema change in relational DB, especially for metadata, we will manage RDB models through its version migration. It's not very flexible, though that's a common feature we must have to support if we use RDB as persistence layer.

Have you happened to consider how schema migrations would work in more detail? I know you mentioned ORMs, but not sure if you had a specific one in mind?

I wonder if it would make sense to make the RDBMS integration the first one to write through metadata service `PUT`s rather than writing directly to the database. I think that's a better model for all integrations going forward, but it would be especially beneficial for RDBMS because it would make DB migrations less painful. What do you think? Would that add too much complexity?

With that in mind, I'm in favor of this as is. I'd be especially excited if it writes through metadata service, but I understand that it might add too much complexity to an already-heavy lift (although if it were simply a matter of lift, I may be able to help do some of that work). 


---

> @dorianj I won't go into the design choice on why we choose graph dbs initially. There are many features with graph db that way more appeal/performant than rds. And I think one of important factor to make Amundsen tick in OSS is the pluggable backend compared to other OSS tool(e.g linkedin wherehows in mysql). This RFC is to add yet another persistent layer using rds(e.g mysql) for those who can't have a graph solution or use a managed graph db solution(e.g aws neptune, neo4j cloud service etc).
> `Makes feature development slower: every new data type would now need to be 100% implemented in both neo4j/Neptune as well as RDBMS.` -- That's not true. Different proxies could support the feature differently as long as they agree with proxy. E.g team behind lyft drove the neo4j, ING drove atlas, and square for neptune.

I agree that pluggability is desirable, and I get that RDBMS is much preferable for most deploys.

But I think that if we commit to pluggability with the current architecture, longer-term we will end up in a state where pluggability is only in theory, not actually usable in practice. We'll end up with a complicated compatibility matrix where only some features work with some back-ends (or even if they are supported, aren't very well used or tested and therefore are not high quality).

I'll use a simple example, let's say that Square and Lyft embark on two projects: Lyft builds lineage, and Square builds push-based ingestion. (ignore the technical details of these for the sake of example)

After those landed, it would now be natural for the Amundsen us to advertise that we support "neo4j, Neptune, column-level lineage, and push-based data ingestion". But for a prospective user, the devil is very much in the details, because likely Lyft would only implement lineage for neo4j, and Square would likely only implement push ingestion for Neptune. The community would be left to implement those features for the other backends. That may happen long after the functionality is shipped.

As we accumulate more features, and more backends, the compatibility matrix will become increasingly complex. If we added other options to the search store, that compatibility matrix would gain another dimension. I think a general expectation is that if we have various backends or integrations, that generally all functionality should be supported in it. I don't think it's a great experience for users if there are many "gotcha's" based on which combination of plugins are supported.

You mentioned wherehows being tied to MySQL as a downfall. datahub supports both Postgres and MySQL and probably any other RDBMS as a core storage engine, because of the abstractions they use (here is [the PR adding Postgres support](https://github.com/linkedin/datahub/pull/1742) -- note that it is only administrative work, not reimplementation on a per-feature basis. Hopefully the use of ORMs in this RFC will allow us to have similar levels of compatibility with both RDBMS as well as possibly document stores). But because RDBMS and graph databases are so different, we aren't able to have a zero-cost abstraction and support both of them to play the same role in an architecture graph.

If we separated the roles of the graph database and the core persistence database, it would be much more possible for us to use abstractions that allow plugins to be implemented once, not on a per-feature basis (e.g. for graph databases, use Gremlin for all queries)


> `I think it will ossify the current architecture wherein databuilder must write directly to the persistence stores (rather than writing through an intermediate services), making it harder to make changes in the future` -- I think starting with RDS like the neptune case in a separate repo helps to make the dev velocity faster. But once the model is stable, we could move it to metadata service and make databuilder calls metadata put api.

I'm not sure what you mean here? Is the proposal landing this RDBMS change, then doing the migration to make databuilder talk to a new metadata put API?

crazy-2020:
Personally speaking, graph db definitely has some advantages in storing metadata. Supporting relational DB as an another backend store option can lower the requirement for use and dev/ops of Amundsen, since some potential users/developers are more familiar with SQL/relational DBs. Migration from Graph db to sole RDB would not only cause big change for dev side but also lead to some work for end users. I am a little bit concerned if current users are willing to do the migration if they are running Amundsen well on graph db.

For the RDB model version migration, it is based on the library 'alembic'. The approach is similar to the one [Airflow](https://github.com/apache/airflow/tree/master/airflow/migrations)  is using now. For the schema change, we need to generate a new model version and alembic provides some good features to upgrade/downgrade to a specific version.

I am okay with where to try to write data to relational db first. I did some prototype work first in databuilder because most concerns of RDB model design comes from ingestion from databuilder. For metaservice change, there would be a big work for proxy code and I would be very grateful if there is any help. 

Definitely, relational DB and graph db have a huge difference which lead to some concerns for future development and maintenance. There are still lots of work needed to do to make them work in the best way, but it would be existing if Amundsen can support multiple types of DBs for workplaces with different DB environment/background.

feng-tao (commented):
some of the previous discussion on the spec could be found in https://github.com/amundsen-io/rfcs/pull/10

feng-tao (approved):

PR #20, opened 2020-12-19
Automated Release Process

mgorsk1:


Golodhros:
I totally support this! We have been talking internally about this for a while.
In my mind, we would have a @ci or @next release process and an official monthly release cadence, so we cover both people looking for only stable releases or user trying to get the latest bug fixes.

mgorsk1:
Thanks @Golodhros , that's another thing I wanted to raise - intermediate releases (something in shape of release candidates). In ING to be as current as possible with latest changes in Amundsen now we rely on git submodules and building Amundsen components ourselves from maste, but having possibility to do `pip install --pre` would come definitely in handy.

rfcs/020-automated-release-process.md
```diff
@@ -0,0 +1,84 @@
...
+- periodical releases every month
+- pre-releases with each change added to master
+
+## Reference-level Explanation (aka Technical Details)
+
+The plan for each release option is:
+
+### Continuous Pre-releases
+
+- Implement a new github action called `Pre-releases` that will trigger on every PR merged into master
```
mgorsk1, line 34: maybe to reduce no of pre-releases we could do them like monthly but with weekly granularity ?

mgorsk1:
slightly off-topic but releasing alongside docker images could also be useful and made onboarding Amundsen even easier

rfcs/000-support-rdb-models.md
```diff
@@ -0,0 +1,122 @@
+- Feature Name: Support Relational Database Models
```
mgorsk1, line 1: is it correct to have this submission under this PR ?

rfcs/020-automated-release-process.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Guide-level Explanation (aka Product Details)
+
+Having enforced naming convention for MRs we can leverage this to automatically categorize MRs and render them in
+changelog. By using automated github actions we can remove manual step from release process (releasing, rendering changelog)
+and replace it with three operations:
+
+- creating a git tag triggers a release
+- periodical releases every month
+- pre-releases with each change added to master
```
mgorsk1, line 26: my feeling is that'd be too granular (as explained in another comment)

rfcs/000-support-rdb-models.md
Golodhros: Oops, something went wrong!

rfcs/020-automated-release-process.md
Golodhros: My thinking here is to cover this use case:

1. User reaches out about a bug
2. We fix it, merge it into master
3. We need to trigger a new release and ask user to update

Instead, we would just fix it and get it into master, asking the user to use the prerelease to get the latest

If all is automatic for us, I don't see any problem. Another thing is if we can't do it automatically, that would be tedious!

feng-tao:
@allisonsuarez are we ok to merge this RFC given all the work you did is merged?

allisonsuarez (approved):

PR #22, opened 2021-02-04
GCP Data Proxy

mgorsk1:

### DIFF
--- a/rfcs/000-gcp-data-catalog-proxy.md
+++ b/rfcs/000-gcp-data-catalog-proxy.md
@@ -1,0 +1,67 @@
+- Feature Name: GCP Data Catalog Proxy Support
+- Start Date: 2021-02-04
+- RFC PR: [amundsen-io/rfcs#](https://github.com/amundsen-io/rfcs/pull/)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
+
+# GCP Data Catalog Proxy Support
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


feng-tao:
Does ING use GCP data catalog + atlas together?

feng-tao:
ping @mgorsk1 , the RFC makes sense though I don't have much context from GCP catalog

mgorsk1:
We've made a cloud PoC some time ago while prepping to cloud migration. Part of this PoC was to test DC instead of Atlas as Amundsen Proxy and we've developed basic version of proxies for both tables and dashboards. The ruling on what our approach we take is still pending, there are some concerns re data security in cloud but I think for metadata it should be fine.

As for using DC with Atlas - it's possible to sync data from Atlas to DC by using one of official connectors: https://github.com/GoogleCloudPlatform/datacatalog-connectors and work well (we've tested with tableau, hive, postgresql and mysqk). They cover syncing metadata from 13 db and 2 dashboarding systems.

MrwanBaghdad:
+1 for integration with GCP and skipping the data builder. However, one concern is how to transform and enrich the table's metadata. One use case that we don't want to expose some tables i.e (depending on policy tags, columns, naming patterns..etc) in other words the level of control we already have using the databuilder.

kristenarmes:
Amundsen is being archived and is no longer actively maintained. As part of this, we're closing all open pull requests, including this one.

Thank you for taking the time to contribute — we appreciate the effort you put into this change, even though we're unable to merge it. If you'd like to continue this work, feel free to fork the repository.

PR #23, opened 2021-02-18
Add MySQL as a data sink in the databuilder

crazy-2020:


rfcs/000-support-mysql-as-backend-store.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Guide-level Explanation (aka Product Details)
+
+With the new serializer, data loader, publisher, users could invoke them in ETL job of databuilder if they are using MySQL as the backend store.
+
+## UI/UX-level Explanation
+
+N/A
+
+## Reference-level Explanation (aka Technical Details)
```
feng-tao, line 24: 1. I think the mysql initialize could be done on the metadata side (e.g either during metadata startup, or provide a cli) per discussion.
2. Should we setup a range dependency which will always use rds latest schema version?
3. I wonder whether you have any prototype and see how long it takes to load metadata to mysql for given N records?

crazy-2020: 1. yes, the db init will be done by metadata service.
2. We will probably have to specify the version range of rds package for databuilder to grab the latest one.
3. The metadata publish job(not including extraction, csv loader job) will take 8mins to load 200K records into mysql. (Given CPU: 8, Memory: 8GB, publish_transaction_size: 500, local communication)

feng-tao: thanks

feng-tao (approved):
lgtm

PR #25, opened 2021-02-23
Lineage Stage 0

allisonsuarez:
Closed #24 in favor of this one

This is an RFC describing the first iteration of Lineage related work.


allisonsuarez:
@danwom use this one instead

rfcs/025-lineage-stage-0.md
```diff
@@ -0,0 +1,149 @@
+- Feature Name: lineage_stage_0
+- Start Date: 2021-02-22
+- RFC PR: [amundsen-io/rfcs#24](https://github.com/amundsen-io/rfcs/pull/24)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000)
+# Amundsen Lineage - Stage 0
+
+## Summary
+
+
+Currently Amundsen doesn't have a way of surfacing lineage information for tables and columns. The idea for this first iteration is to have a way to show upstream and downstream tables and columns to users through the Table Details page so they can explore the current resource's lineage as well as navigate to related resources in Amundsen.
```
feng-tao, line 10: FYI, actually we could use programmatic description to surface the lineage . I think what it is lacking, it is a graph UI to surface the lineage intuitively.

rfcs/025-lineage-stage-0.md
```diff
@@ -0,0 +1,149 @@
...
+> Explaining the feature largely in terms of examples.
+> Explaining how Amundsen users should think about the feature, and how it should impact the way they use Amundsen. It should explain the impact as concretely as possible.
+> If applicable, provide deprecation warnings, or migration guidance.
+> For implementation-oriented RFCs, this section should focus on how maintainers should think about the change, and give examples of its concrete impact. For policy RFCs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms.
+### New Concepts
+- Lineage:
+- Upstream:
+- Downstream:
+
+Those implementing Amundsen should keep in mind that this feature is meant to provide them with a way to surface their existing lineage data by calling the service containing that data from the metadata service. This iteration won't provide a model to persist lineage on neo4j, but rather a gateway to lineage data so it can be included on lineage API responses to displayed in frontend. It is also important to understand that the feature will be disabled by default and can be enable through configuration.
```
feng-tao, line 31: are you going to put the service call (from Amundsen to CW or 3rd party internal service) in Lyft internal? 

Also wouldn't https://github.com/amundsen-io/amundsencommon/pull/90 surface as the model?

rfcs/025-lineage-stage-0.md
```diff
@@ -0,0 +1,149 @@
...
+```
+{
+  “key”: “current_table_key”,
+  “direction”: “upstream”
+  “lineage_entities_upstream”: [
+    {
+      “table”: “table_key1”,
+      “level”: 1,
+      "source": “hive”,
+      “badges”: [“coco”, “beta”],
```
feng-tao, line 66: is https://github.com/amundsen-io/amundsenmetadatalibrary/pull/259/files matched this called? or the pr is still WIP?

rfcs/025-lineage-stage-0.md
allisonsuarez: Yeah, I mean in this iteration we won't be adding any relations or lineage information to the neo4j graph

rfcs/025-lineage-stage-0.md
allisonsuarez: the metadata PR is still WIP but it will be the same

feng-tao:
the png in https://github.com/amundsen-io/rfcs/blob/59bd5b98c15e823fcf1757fcd5ff97790e215dae/rfcs/025-lineage-stage-0.md doesn't seem to work

feng-tao:
I will take a look tonight

rfcs/025-lineage-stage-0.md
```diff
@@ -0,0 +1,153 @@
...
+Additionally we will add lineage information at the column level, viewable by expanding column metadata.
+
+These features will only appear when the lineage feature is enabled.
+
+## Reference-level Explanation (aka Technical Details)
+### Architecture
+
+![Lineage Stage 0 Architecture](assets/025/lineage-arch.png)
+
+Implementing this feature will require defining a Lienage API on the metadata service for Tables and Columns. When the API is called it will make calls to neo4j and whatever the source of lineage data is. An interface needs to be created to interact with an implementer's lineage service in a generic way. The data from the calls to these services will be put together to form the lineage response as defined below.
```
verdan, line 47: Would this generic API call be made to any of the metadata proxies? or is this simply a configurable function which users will be able to extend themselves? 
Suppose a user has a proxy set to `Atlas`. For Stage 0, will this endpoint call a function inside the atlas_proxy?

verdan (approved):
This looks good to me for Stage 0.

rfcs/025-lineage-stage-0.md
allisonsuarez: Yeah on the Base Proxy class I added a get_lineage method that can be implemented and will be called for any proxy when the endpoint is hit https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/metadata_service/proxy/base_proxy.py#L160

feng-tao:
@danwom @allisonsuarez here is the preview (https://github.com/amundsen-io/rfcs/blob/59bd5b98c15e823fcf1757fcd5ff97790e215dae/rfcs/025-lineage-stage-0.md) . Could you update and fix the broken image links? thanks

rfcs/025-lineage-stage-0.md
```diff
@@ -29,11 +29,11 @@ Those implementing Amundsen should keep in mind that this feature is meant to pr
 
 ## UI/UX-level Explanation
 
-![Table Lineage Preview](assets/025/table-lineage-preview.png)
+![Table Lineage Preview](../assets/025/table-lineage-preview.png)
 
 We will add two additional tabs to the `Table Details` page, `Upstream` and `Downstream`. Each tab will contain a list of tables from which data is inherited or consumed. This allows users view a table's lineage in a very simple manner.
 
-![Column Lineage Preview](assets/025/column-lineage-preview.png)
+![Column Lineage Preview](../assets/025/column-lineage-preview.png)
```
feng-tao, line 36: any reason we don't go with lineage tab with column and dashboard, then have upstream/downstream subtab underneath?

danwom: I think there's a preference to avoid nested tabs when possible. At some point if we have too many tabs then we can reconsider nesting or other options.

mgorsk1: For me it feels weird to break it into two tabs. lineage is, after all, a big picture of data flow - so would make sense to have it in one place. It'd make it easier to analyze it for end user.

rfcs/025-lineage-stage-0.md
mgorsk1: So the idea is that with this method any lineage provider (like Marquez, spline or open lineage) can be supported, right?

rfcs/025-lineage-stage-0.md
feng-tao: yeah, +1 on @mgorsk1 , I felt the same to have single lineage tab with nested tab for upstream/downstream (not sure how complexity it changes for FE implementation). @danwom @allisonsuarez could you share more about the details of the preferences?

rfcs/025-lineage-stage-0.md
feng-tao: @mgorsk1 I think this is just step 0 which we should bring the backend model (or I assume we will bring) with ingestion in step 1/2. The reason on why we are doing this is because Lyft still uses a 3rd party vendor lineage service . It is easier to for the implementation in these kinds of steps by focusing on FE first with backend proxy; then build a backend model/ ingestion with push mechanism in a subsequent step.

rfcs/025-lineage-stage-0.md
verdan: +1 for not using the nested tabs where possible.

For Stage 0, it will be a list of items for each direction, which IMHO is okay to be placed in different tabs. But yes, for the next milestone, where we'll have a graph/chart of the complete lineage (and not the list of tables), that must be placed in one view under 1 tab to have a clear view of the complete lineage.

rfcs/025-lineage-stage-0.md
allisonsuarez: Yeah we will hopefully add a backend model for lineage in the future so that we can query the db directly for lineage rather than making ad hoc calls to a provider from metadata. In that case we would have to add support for lineage extraction on databuilder.

feng-tao (approved):
lgtm for stage 0

rfcs/025-lineage-stage-0.md
mgorsk1: I am not really advocating for nested tabs as this sounds more like implementation preference rather than UX question. Is nesting the only way this could be implemented in single tab ? I would even see something like single list (sorted by upstream and then downstream) with table names and icon/abbreviation if the table is upstream or downstream.

rfcs/025-lineage-stage-0.md
mgorsk1: Clear, for me support for external service is more than fine, just wanted to make sure I got it right.

PR #29, opened 2021-03-22
Table and Dashboard Notices

Golodhros:
Resource Notices allows Amundsen administrators to show informational, warning and alert messages related to the different resources (tables, dashboards, people) we expose in Amundsen.

Here are some visuals of them:

Info:
redshift_dimension_rides_-_Amundsen_Table_Details

Warning:
Amundsen_-_Data_Discovery_Portal

Here is a draft WIP PR for this feature: https://github.com/amundsen-io/amundsenfrontendlibrary/pull/957


rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+
+# Resource Notices
+
+## Summary
+
+Resource Notices allows Amundsen administrators to show informational, warning and alert messages related to the different resources (tables, dashboards, people) we expose in Amundsen
+
+## Motivation
+
+This feature will help administrators to show messages related to deprecation, updates (or lack of), and informational messages that are related to specific resources. In Lyft's case, we are using them to point final users to a improved versions of current surfaced tables, so final users employ more trusted data.
```
allisonsuarez, line 14: an* improved

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Guide-level Explanation (aka Product Details)
+
+An Amundser final user can see a notice in the table/dashboard/user detail page whenever his system admin adds one related to it.
```
allisonsuarez, line 18: Amundsen*

allisonsuarez: also use their instead of his

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Unresolved questions
+
+We think this proposal is a good starting point. It might be lacking in different ways, but we don't know yet in which ways. That's why we think this would be only the first iteration.
+
+## Future possibilities
+
+Out of scope and pending of feedback for this stage will be:
+
+- A different way of storing the notices (JSON file, as part of the graph, UI feeded)
```
allisonsuarez, line 86: fed

allisonsuarez (approved):
Looks great to me for an initial iteration of this feature, just left a few small nits

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+
+It is not a breaking change, there won't be any costs for older versions to support this.
+
+## Drawbacks
+
+- This could not be scalable, as maybe users want to have too many notifications. It doesn't mean this work won't be a good start, but it could be not enough. TBD
+- With this approach, in order to add/update/remove a notification, admins would need to make a new deploy of the application. A better approach would be if we use a UI, so a proper feature flag system would be much better for this.
+- Notices are in a 1:1 basis. If admins wanted to deprecate a whole schema or dashboard product, they would need to create a notice for each table/dashboard.
+
+## Alternatives
```
dorianj, line 68: You already mentioned this in future possibilities, but I think the primary alternative to consider is sourcing this stuff from the graph itself. I think that would be preferable to feature flagging systems, since it's integrated into the graph and not tacked-on (unless i'm missing exactly how the feature flag system would be used here)

agreed to leave it out-of-scope for now

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+
+The design we propose is based in the following:
+
+- Using our current frontend configuration files (config-utils.ts, config-default.ts, config-custom.ts and config-types.ts)
+- Allowing a three level "severity" schema for the alerts (info, warning, and alert)
+- Allowing its use in "tables" and "dashboards" (we could add it to people as well, but we'll need designs for it)
+
+What we are not supporting for now:
+
+- Multiple notifications per resource
```
dorianj, line 54: I think we should support this from the get-go: it will make it far easier to have multiple sources of notices in the future if everything is array-aware in the UI. Thoughts?

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+## Drawbacks
+
+- This could not be scalable, as maybe users want to have too many notifications. It doesn't mean this work won't be a good start, but it could be not enough. TBD
+- With this approach, in order to add/update/remove a notification, admins would need to make a new deploy of the application. A better approach would be if we use a UI, so a proper feature flag system would be much better for this.
+- Notices are in a 1:1 basis. If admins wanted to deprecate a whole schema or dashboard product, they would need to create a notice for each table/dashboard.
+
+## Alternatives
+
+- Integrating a proper feature flag system and feed the configuration from it. This would be ideal not just for this feature but for many others. It is also more complex and costly.
+- Using a JSON file to feed the notices was also considered. Given that we are not happy with the current configuration options (JS config + Python configuration), adding yet another one didn't feel like a good move.
```
dorianj, line 71: agreed

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+- If we don't do anything, we won't be able to show this information in a highlighted way (we would probably add this as a programmatic description, making it less visible)
+
+## Prior art
+
+Nothing that we know of
+
+## Unresolved questions
+
+We think this proposal is a good starting point. It might be lacking in different ways, but we don't know yet in which ways. That's why we think this would be only the first iteration.
```
dorianj, line 80: i think it's a good starting point

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+
+A notice is a small box with an icon and a message that can contain HTML markup (like links and bolded text).
+
+These will come in three flavors:
+
+- Informational: Marked with a blue "i" icon on the right side
+- Warning: Marked with an orange exclamation mark icon on the right side
+- Alert: Marked with a red exclamation mark icon on the right side
+
+For example, if company X wants to deprecate the use of one table, they can opt to add a new notice in their configuration file:
```
dorianj, line 28: how does this config change for table / dashboard / etc? Possibly: the key be the object URI?

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+
+- Using our current frontend configuration files (config-utils.ts, config-default.ts, config-custom.ts, and config-types.ts)
+- Allowing a three-level "severity" schema for the alerts (info, warning, and alert)
+- Allowing its use in "tables" and "dashboards" (we could add it to people as well, but we'll need designs for it)
+
+What we are not supporting for now:
+
+- Multiple notifications per resource
+- Loading of external files (like a JSON)
+- Application of a single notice to multiple resources
```
dorianj, line 56: yep, seems like a fine limitation for now. could change the keys to be native regex objects in the future if we wanted.

dorianj (commented):
I'm +1 on this. I do think that supporting multiple notices will make it easier to transition in the future, so that's my only change request, but I could also be convinced that it's not hard to change in the future and so we should ignore it for now

rfcs/000-resource-notices.md
Golodhros: You are probably right!

I guess the benefit of the feature flags would be to allow anybody to create/update/remove notifications.

rfcs/000-resource-notices.md
Golodhros: The main problem we see is how do we limit the notifications we show.

Do we set an arbitrary limit like three? Or should we add a "see more" link that expands? Or a modal?

I would be OK to add it if we limit them with an arbitrary threshold, but the rest seem too much thinking in advance for us right now!

rfcs/000-resource-notices.md
Golodhros: Yeah, we currently have different configs for the different resources, so we will be adding a "notices" key in each one of those.

rfcs/000-resource-notices.md
dorianj: I sorta think it's the responsibility of the config writers to restrain themselves in how many they include. that said, hard limiting to 2 or 3 is eminently reasonable.

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+These will come in three flavors:
+
+- Informational: Marked with a blue "i" icon on the right side
+- Warning: Marked with an orange exclamation mark icon on the right side
+- Alert: Marked with a red exclamation mark icon on the right side
+
+For example, if company X wants to deprecate the use of one table, they can opt to add a new notice in their configuration file:
+
+```
+    ... Table Configuration
```
feng-tao, line 31: does it only support config based notice instead of dynamic notice? E.g could the feature be used to show a table has data delay / quality issue which will require realtime metadata ingestion?

rfcs/000-resource-notices.md
dorianj: which config key in `AppConfig` is for each of them? I'm also OK with deferring this discussion to the PR, that would be fine

rfcs/000-resource-notices.md
dorianj: this proposal is static only -- there is discussion / note that adding it to the graph (and thus realtime) is currently out-of-scope, but compatible with this change (there are use cases for each of these)

rfcs/000-resource-notices.md
slavierouse: I suggest using recency, i.e. we show up to the 3 latest notices

rfcs/000-resource-notices.md
slavierouse: +1

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Drawbacks
+
+- This implementation could not be scalable, as maybe users want to have too many notifications. It doesn't mean this work won't be a good start, but it could be not enough. TBD
+- With this approach, to add/update/remove a notification, admins would need to make a new deployment of the application. A better approach would be if we use a UI, so a proper feature flag system will be much better for this.
+- Notices are on a 1:1 basis. If admins wanted to deprecate a whole schema or dashboard product, they would need to create a notice for each table/dashboard.
+
+## Alternatives
+
+- Integrating a proper feature flag system and feed the configuration from it. This would be ideal not just for this feature but for many others. It is also more complex and costly.
```
slavierouse, line 70: Is feature flagging the right type of system? imo feature flags means turning on or off certain features for a population, for example for a/b testing. This seems more like administrator UI or administrator privileges

rfcs/000-resource-notices.md
dorianj: the notices config doesn't have dates associated with them... so I think priority sort is only thing possible without changing that spec?

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Prior art
+
+Nothing that we know of
+
+## Unresolved questions
+
+We think this proposal is a good starting point. It might be lacking in different ways, but we don't know yet in which ways. That's why we think this would be only the first iteration.
+
+## Future possibilities
```
slavierouse, line 82: Suggesting the following as roadmap:

- MVP: 1 non-dismissable notice per resource | notice data stored in ConfigUtils
- v1: up to 3 dismissible notices per resource shown, ranked by recency | notice data stored in neo4j 
- v2: form for creating notices | access control so only owners or admins can add or remove notices on their resources from the UI

rfcs/000-resource-notices.md
Golodhros: Only config based for now.
That' a nice use case, we should consider it for the next phases of this feature!

rfcs/000-resource-notices.md
Golodhros: True, maybe feature flags won't be the best match, but they would allows us to change them on the fly.

Yeah, an admin UI would be a better fit, but I don't know if we have more use cases for it, worth to keep it in mind!

rfcs/000-resource-notices.md
Golodhros: Adding those, thanks!

rfcs/000-resource-notices.md
Golodhros: is a dynamic one [ResourceType.table], [ResourceType.dashboard] and so on

rfcs/000-resource-notices.md
Golodhros: True, there are no dates and we cannot ensure the position of elements on the object as we render them.

mgorsk1: I am also in favor of `no-limits`. True, users might create a lot of them in graph, but it is the same case with Programmatic Descriptions. It's up to end users to say how many of them are required. We could enforce sorting, so the errors are always on top, later there are warnings and info are on the bottom.

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,94 @@
...
+        "<SCHEMANAME>.<TABLENAME>": {
+          severity: NoticeSeverity.ALERT,
+          message: `<span>This table is deprecated, please use <a href="<LINKTONEWTABLEDETAILPAGE>">this new table</a> instead.</span>`,
+        },
+    },
+```
+
+This code will show a notice with a red exclamation icon whenever a final user visits the table's Table Detail page.
+
+This feature's ultimate goal is to allow Amundsen administrators to point their users to more trusted/higher quality data without removing the old references.
```
mgorsk1, line 42: higher quality would be just one use-case I think, this can easily be used for
* recommending different tables (`more like this` feature or `other users also queried`)
* notifying about failed data quality checks
* notifying about maintenance on the tables (like it's being repartitioned and it will take X hours)

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,94 @@
...
+Out of scope and pending feedback for this stage will be:
+
+- A different way of storing the notices (JSON file, as part of the graph, UI fed)
+- Multiple notifications at once and how we would limit those
+- Notifications in the User pages
+
+A proposed roadmap is:
+
+1. MVP: 1 non-dismissable notice per resource | notice data stored in ConfigUtils
+2. v1: up to 3 dismissible notices per resource shown, ranked by recency | notice data stored in neo4j
```
mgorsk1, line 93: 1. maybe ranked by severity and then recency ?
2. another idea would be also to have TTL on the notices (when adding them we could have `expiration_date` optional field after which they would not be displayed)

rfcs/000-resource-notices.md
```diff
@@ -0,0 +1,94 @@
...
+3. v2: form for creating notices | access control so only owners or admins can add or remove notices on their resources from the UI
```
mgorsk1, line 94: maybe we should also consider api for dynamically generating notices. I can imagine situation where some ETL pipeline validates data quality and based on it sends a notice `this table has not passed latest quality check`

rfcs/000-resource-notices.md
mgorsk1: Amundser and Amundsenmadamme 😂

rfcs/000-resource-notices.md
Golodhros: That's a good point @mgorsk1. I also think sorting by severity makes sense.

rfcs/000-resource-notices.md
Golodhros: Those are great examples, thanks!

rfcs/000-resource-notices.md
Golodhros: Great ideas, adding them!

rfcs/000-resource-notices.md
Golodhros: Good call, adding it!

Golodhros:
Given there is no opposition to this plan, moving this into FCP!

rfcs/000-resource-notices.md
mgorsk1: One more thought - what is the ballpark of notices we can add to config before performance starts to decay ? Would having 1k+ notices be reasonable ?

Golodhros: Using the configuration it is a O(1) operation, so there is no issue in the number. It will look terrible though and it would break the balance on the design though.

mgorsk1: ok thanks, I meant 1 notice per table & add notices for 1k tables. In our case we have over 10k tables in Amundsen and a lot of them could use deprecation/pointing out to newer version.

mgorsk1 (approved):

rfcs/000-resource-notices.md
Golodhros: Oh, I see. In theory is fine, just a PITA to add! :)

rfcs/029-resource-notices.md
```diff
@@ -0,0 +1,101 @@
...
+
+This code will show a notice with a red exclamation icon whenever a final user visits the table's Table Detail page.
+
+This feature's ultimate goal is to allow Amundsen administrators to point their users to more trusted/higher quality data without removing the old references.
+
+## Reference-level Explanation (aka Technical Details)
+
+The design we propose is based on the following:
+
+- Using our current frontend configuration files (config-utils.ts, config-default.ts, config-custom.ts, and config-types.ts)
```
verdan, line 48: @Golodhros, The only thing I'd propose here is to use the Python/Flask config instead of JS config for this. i.e., config.py

Reasons:
- This sounds like more a metadata feature rather than a UI thing. 
- Not everyone is doing the customization of the frontend in the community. 
- For most of us, we are already doing the modification in our flask configs, so relatively easy to keep the maximum changes in one place. I'd try to minimize the places of configurations. 
- We won't need to change the frontend code in the future, as we are already serving the notices from the backend.

Golodhros: I agree with you in that we should minimize the places where we have configurations, and I would love us to move into a feature flag system plus an admin UI to keep all this static configs.

I think the right place for this (for now) is the JS configuration because:
- The notices have a severity attribute that affects the looks, and we can use TypeScript for making sure they are one of the options we allow
- For now, it is not a metadata feature as we don't get the input from a third party source
- With the current state of things, both changes on the Flask config and the JS config would need a PR in the same repo, so it is basically the same without the type safety

I think that with the changes planned on the roadmap for this feature, we will end up in a place that is handy for everyone, but just for now bear with me!

verdan: sounds good to me!

verdan (approved):
nice work...

rfcs/029-resource-notices.md
```diff
@@ -0,0 +1,101 @@
+- Feature Name: Resource Notices
```
jcarey03, line 1: Another common term I've seen and used for this type of thing is `bulletins`.

rfcs/029-resource-notices.md
```diff
@@ -0,0 +1,101 @@
...
+
+Resource Notices allows Amundsen administrators to show informational, warning, and alert messages related to the different resources (tables, dashboards, people) we expose in Amundsen.
+
+## Motivation
+
+This feature will help administrators show messages related to deprecation, updates (or lack of), and informational messages related to specific resources. In Lyft's case, we use them to point final users to improved versions of current surfaced tables, so final users employ more trusted data.
+
+## Guide-level Explanation (aka Product Details)
+
+An Amundsen final user can see a notice in the table/dashboard/user detail page whenever their system admin adds one related to it.
```
jcarey03, line 18: Will there be an API to create notices? For example, a side-effect of a SEV is the automatic creation of a notice for the impacted resources.

rfcs/029-resource-notices.md
```diff
@@ -0,0 +1,101 @@
...
+
+A notice is a small box with an icon and a message containing HTML markup (like links and bolded text).
+
+These will come in three flavors:
+
+- Informational: Marked with a blue "i" icon on the right side
+- Warning: Marked with an orange exclamation mark icon on the right side
+- Alert: Marked with a red exclamation mark icon on the right side
```
jcarey03, line 26: Can you give an example or some semantics around each notice level?

rfcs/029-resource-notices.md
```diff
@@ -0,0 +1,101 @@
...
+
+The design we propose is based on the following:
+
+- Using our current frontend configuration files (config-utils.ts, config-default.ts, config-custom.ts, and config-types.ts)
+- Allowing a three-level "severity" schema for the alerts (info, warning, and alert)
+- Allowing its use in "tables" and "dashboards" (we could add it to people as well, but we'll need designs for it)
+
+What we are not supporting for now:
+
+- Multiple notifications per resource
```
jcarey03, line 54: Curious about why multiple notifications will not be supported? Is it a lack of use case or technical challenges or to keep feature scope smaller?

rfcs/029-resource-notices.md
```diff
@@ -0,0 +1,101 @@
...
+- Loading of external files (like a JSON)
+- Application of a single notice to multiple resources
+
+In the implementation, we'll check the notices object whenever we visit the resource detail page, showing any notification if present.
+
+It is not a breaking change, there won't be any costs for older versions to support this.
+
+## Drawbacks
+
+- This implementation could not be scalable, as maybe users want to have too many notifications. It doesn't mean this work won't be a good start, but it could be not enough. TBD
```
jcarey03, line 64: This matches my intuition. The proposal is quite limiting. A good start, but I think having a human in the loop and only have one notification really limits this proposal's utility.

rfcs/029-resource-notices.md
```diff
@@ -0,0 +1,101 @@
...
+
+A proposed roadmap is:
+
+1. MVP: 1 non-dismissable notice per resource | notice data stored in ConfigUtils (current implementation in this [PR](https://github.com/amundsen-io/amundsenfrontendlibrary/pull/957))
+
+2. v1: up to 3 dismissible notices per resource shown, ranked by severity and recency (or expiration date via TTL) | notice data stored in neo4j
+   Some use case examples for v1 are:
+
+- Dynamically recommending different tables (more like this feature or other users also queried)
+- Notifying about failed data quality checks
```
jcarey03, line 98: Will this still require an admin to create the notice?

rfcs/029-resource-notices.md
Golodhros: It wasn't anything we were thinking at the beginning, but it has been a common suggestion so we added it on the phase 3 of the roadmap below.

rfcs/029-resource-notices.md
Golodhros: I think that will be up to the Amundsen admins to decide, we ourselves were doubting what to use for the ones we need to place right now.

rfcs/029-resource-notices.md
Golodhros: To keep the feature scope small. We added multiple notices to the v1 in the roadmap.

rfcs/029-resource-notices.md
Golodhros: Somebody that could access the neo4j graph for sure.

rfcs/029-resource-notices.md
feng-tao: +1 on Jason @jcarey03 I think we may want to have API driven to set notice (at least in long run).

rfcs/029-resource-notices.md
mgorsk1: I think for da checks it could be api access as well

rfcs/029-resource-notices.md
jcarey03: Let me ask this differently, how did we decide on three categories? Do we expect this list to change over time?

Golodhros: We got feedback that having more types than the initial one (alert) would be useful. We agreed and introduced the basic severity pack (3 severities).

It could grow larger, maybe to 5 or 7, and it would be easy and backwards compatible if we decide to do so in the future.

PR #31, opened 2021-03-30
Add new RFC for monorepo

dorianj:
This is a continuation / down-scoping of https://github.com/amundsen-io/rfcs/pull/30. From that discussion, it seems that there is clear support for moving to a mono repo. We'd like to figure out the details of that change and get that rolling and defer the other things that were bundled in the prior RFC.

### DIFF
--- a/rfcs/031-monorepo.md
+++ b/rfcs/031-monorepo.md
@@ -41,7 +41,17 @@ and many more... This change will put all the packages in just one place, making
 
 1. Code Duplication: From requirements to config files, models, helper functions, exceptions handling, CI/CD pipelines, Docker files, and other repository management like licenses, PR/Issues templates, etc., is pretty much the same in all these microservices. Change in one repository deviates that repository from others at the moment. This change will move all the above into a single repository making it easy to change anything without keeping track of multiple repositories, and redundant code.
 
-Having one repo doesn't preclude us having incubator repositories, they can be handled similarly as today.
+
+### Contrib/incubating packages
+
+From time-to-time, the Amundsen project creates third party packages to allow substantial new additions to be developed by non-maintainers. The monorepo per se doesn't preclude us having incubator repositories, they can be handled similarly as today.
+
+We currently have two approaches for proxy development:
+
+1. 'old' proxies like neo4j, elasticsearch or atlas are kept directly in amundsenmetadata and amundsensearch respectively
+1. 'new' proxies like rds or gremlin are separate git repositories.
+
+Our intention in the monorepo would be to keep incubating repos separate only until they are stable, upon which they should be merged into the main repo. Any officially-supported proxies should get CR by maintainers, not exist outside of that process.
 
 
 ## Transition Path
@@ -55,6 +65,8 @@ We create a new branch on the `amundsen` repo, named `main`. All work that follo
 Raw code import and CI. We import all of the code as a snapshot into the repository.
 
 At this point, we migrate the CI/CD pipeline, and automated release mechanisms, from the consituent repositories to the main repository. They will publish packages using a `-beta` postfix.
+
+We will merge the codeowners files at the same time here.
 
 ### Phase 3:
 We freeze code in all the submodules: no new PRs in. Existing PRs will have 1 week to land. We will assist authors in getting nearly-mergable PRs landed. For PRs that don't make it, we will provide instructions to port their patches to the new repo, if they choose to re-submit.
@@ -73,7 +85,6 @@ We mark the `main` branch on GitHub as the main branch, so that new pulls use th
 
 ### Phase 5:
 After 1 year, we will remove all of the sub-repos, as well as the `master` branch on the Amundsen repository.
-
 
 ## How We Communicate This
 
--- a/rfcs/031-monorepo.md
+++ b/rfcs/031-monorepo.md
@@ -7,12 +7,12 @@
 
 ## Summary
 
-At the moment, Amnundsen project is divided into 6 repositories: 
+At the moment, Amnundsen project is divided into 6 repositories:
 	- amundsen
 	- amundsenfrontendlibrary
 	- amundsenmetadatalibrary
 	- amundsensearchlibrary
-	- amundsencommon 
+	- amundsencommon
 	- amudnsendatabuilder
 
 This RFC proposes merging all of these repositories into a single repo.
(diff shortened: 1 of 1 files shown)


Golodhros:
Thanks a lot! 
This looks like the perfect step forward!

rfcs/031-monorepo.md
```diff
@@ -0,0 +1,100 @@
...
+
+At the moment, Amnundsen project is divided into 6 repositories: 
+	- amundsen
+	- amundsenfrontendlibrary
+	- amundsenmetadatalibrary
+	- amundsensearchlibrary
+	- amundsencommon 
+	- amudnsendatabuilder
+
+This RFC proposes merging all of these repositories into a single repo.
```
friendtocephalopods, line 18: Any reason not to add amundsengremlin to the list as well?

dorianj: Is the intention for `amundsengremlin` to get merged into one of the proxies, or is the long-term intention to leave it as a separate package? I'm not super familiar with the code split there

rfcs/031-monorepo.md
```diff
@@ -0,0 +1,100 @@
...
+
+There will be no change to the built artifacts. Users who use PyPi packages or Docker images will have no change to their deployments. Users with forks or submodule based customization will have to make minor changes to their CI/CD pipelines to pull from the correct directories, but the required changes will be minor.
```
mgorsk1, line 20: Smooth change management, nice !

However, with current structure, when someone is using the submodule approach can freely choose versions (even refs) for each product separately (I can use frontend in different version than metadata and search). This change somehow limits that capability. It's worth to explicitly mention it in drawbacks.

rfcs/031-monorepo.md
mgorsk1: I think that's a point that could have more attention in this rfc. We currently have two approaches for proxy development:
1. 'old' proxies like neo4j, elasticsearch or atlas are kept directly in amundsenmetadata and amundsensearch respectively
2. 'new' proxies like rds or gremlin are separate git repositories.

Maybe we should unify the approach ? afaik the idea with having proxies in different repositories made contributing easier - you could have a team that would be maintainers only for proxy repo, not core amundsen repos. @feng-tao would be more knowledgeable here

rfcs/031-monorepo.md
```diff
@@ -0,0 +1,100 @@
...
+
+
+## Motivation
+
+The submodule architecture creates the following issues and is making it hard for companies to adopt Amundsen, and to customize it accordingly. It also causes substantial developer toil in the form of having to create multiple small PRs to merge one 
+
+1. Dependency Management: Python packages are pretty much the same for each proxy, which makes it hard to sync across all the above repositories. As a result most of the time, each proxy is running its own version of the dependency.
```
mgorsk1, line 27: Another approach (or used alonside changes from this RFC to strengthen the dependency management even more) would be introducing the pip constraints files https://pip.pypa.io/en/stable/user_guide/#constraints-files

rfcs/031-monorepo.md
```diff
@@ -0,0 +1,100 @@
...
+
+| Package             | amundsenmetadata   | amundsenfrontend  |
+| --------------      |:------------------:| -----------------:|
+| amundsen-common     | >=0.8.1            | ==0.6.0           |
+| flake8              | ==3.5.0            | ==3.8.4           |
+| flake8-tidy-imports | ==1.1.0            | ==4.2.1           |
+
+and many more... This change will put all the packages in just one place, making practically to keep dependencies similar accross verisons, simply because it will be possible to open a single PR that updates multiple packages. Practically speaking this will improve the status quo slightly.
+
+1. Development Efforts: Having multiple repositories makes it really hard to implement a feature. Implementation and testing require efforts to synchronize and then code reviews, and finally, all the PRs across multiple repositories need to land in master at a certain time or at the same time. This change results in faster development, one PR to fix a bug or a feature, no dependency hell as we are facing today, hence attract more contributors.
```
mgorsk1, line 38: just a question - do we have any hard data (like results of questionnaires) to back the validity of those issues ? or is it more the result of interacting with users and observation ?

rfcs/031-monorepo.md
Golodhros: We could make more extensive use of codeowners files for each package, so we keep the ownership clear.

rfcs/031-monorepo.md
dorianj: these are new to me! That's super cool, added.

rfcs/031-monorepo.md
dorianj: We don't -- we did the call for feedback in December or January from community members but only got a handful of detailed responses. This is from our interactions with many community members (both longer-time installs) as well as people new to the project. Both groups are impacted, but in different ways.

rfcs/031-monorepo.md
dorianj: ah, good point, added

rfcs/031-monorepo.md
dorianj: Proposed some language to consider the incubating repos. Open to feedback!

My thought is the separate packages should only be used for incubating, unsupported packages. Once it it officially supported by the project, it should require CR and RFCs from approved maintainers. Otherwise, we will have officially-supported parts of the project that aren't necessarily governed by our normal processes (RFC and CR). If the outside contributor is still making meaningful contributions after the package is landed, we should consider suggesting them as a maintainer (in which case @Golodhros's suggestion of stronger codeowners would make sense -- it might make sense to have that new maintainer start as specialized to the specific area they were previously working on, and if they want to contribute outside of that, we add that on later via codeowner additions)

feng-tao: currently neptune and RDS backend is introduced as part of RFC process. Though I think it is easier for maintain with different package. But I wouldn't call them incubating repos. This is same as Atlas which the engineering team built the whole integration with neo4j while atlas is maintained by @bolkedebruin ING team. Both are officially supported backend in Amundsen for long time.

feng-tao:
as long as different package / artifacts are maintained within single monorepo, I am +1 for the change. But it would be good for @allisonsuarez , @dkunitsk , @jinhyukchang @youngyjd backend ICs from Lyft to take a look and see anything related to existing integrations.

rfcs/031-monorepo.md
dorianj: Yeah, that makes sense.

> Though I think it is easier for maintain with different package

Can you elaborate on what is made easier here? I'd like to capture the benefits of those packages without the overhead of actually having separate repos, if that makes sense.

feng-tao: The RFC here is to propose mono repo with different packages. For RDS, gremlin based, it seemed to be easier to maintain them separately given the people who developed those are not part of the core committer team if that makes sense.  It will help to iterate those features faster without asking existing committer to signoff.

But once the repos/features are mature, we should bring them into same repo each of which to be a new package.

Golodhros (approved):

PR #32, opened 2021-03-31
Adds an RFC to implement lineage backend

verdan:

### DIFF
--- a/rfcs/032-lineage-backend.md
+++ b/rfcs/032-lineage-backend.md
@@ -7,32 +7,37 @@
 
 ## Summary
 
-Currently, we link to the 3rd Party lineage tool in order to show table and column lineage for tables in Amundsen. 
+Currently, we link to the 3rd Party lineage tool in order to show table and column lineage for tables in Amundsen.
 The goal of this project is to have a lineage graph that is native to Amundsen so that users can find table and column lineage in one platform.
 
 ## Motivation
 
-At the moment we are relying completely on any 3rd party lineage tool to show the table and column lineage in Amundsen. 
+At the moment we are relying completely on any 3rd party lineage tool to show the table and column lineage in Amundsen.
 Not everyone has access to such a 3rd party tool and want to have lineage information available natively in Amundsen.
 
 ref: https://github.com/amundsen-io/rfcs/blob/master/rfcs/025-lineage-stage-0.md
 
 ## Guide-level Explanation (aka Product Details)
 
-No new concepts/definitions are being introduced as a part of this RFC. 
+No new concepts/definitions are being introduced as a part of this RFC.
 
-Databuilder already has the table lineage model, which creates an upstream/downstream relation to adding to the Neo4j graph. 
+This implementation will only cover Neo4j. Please see Future Work for what was descoped.
+
+Databuilder already has the table lineage model, which creates an upstream/downstream relation to adding to the Neo4j graph.
 Column lineage model however still needs to be developed as a part of this RFC.
 
-Important to note: We will not deprecate the ability to show lineage graph/information from 3rd party lineage tools 
-(which is implemented in Stage 0), hence this implementation will be compatible with the existing functionality. 
+More specifically, most of the existing `TableLineage` model will be extracted into `BaseLineage`, from which `TableLineage` and the new `ColumnLineage` will inherit from. These will use `uri`s instead of properties as identifiers and constructors, but otherwise will largely be the same as the current `TableLineage`.
 
+This RFC does not specify databuilder extractors at the moment. Please see Future Work.
+
+Important to note: We will not deprecate the ability to show lineage graph/information from 3rd party lineage tools
+(which is implemented in Stage 0), hence this implementation will be compatible with the existing functionality.
 
 
 ## UI/UX-level Explanation
 
-No UI changes required as a part of this RFC. 
-The implementation should work and support the existing UI, and will not break any of the existing UI. 
+No UI changes required as a part of this RFC.
+The implementation should work and support the existing UI, and will not break any of the existing UI.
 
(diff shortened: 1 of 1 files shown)


rfcs/032-lineage-backend.md
```diff
@@ -0,0 +1,97 @@
...
+Not everyone has access to such a 3rd party tool and want to have lineage information available natively in Amundsen.
+
+ref: https://github.com/amundsen-io/rfcs/blob/master/rfcs/025-lineage-stage-0.md
+
+## Guide-level Explanation (aka Product Details)
+
+No new concepts/definitions are being introduced as a part of this RFC. 
+
+Databuilder already has the table lineage model, which creates an upstream/downstream relation to adding to the Neo4j graph. 
+Column lineage model however still needs to be developed as a part of this RFC.
```
dorianj, line 25: It would be reasonable to define this in the RFC. That said, given there's already table lineage, and the metadata response formats are defined, I think the solution space is small enough it's probably OK to figure it out in the implementation

dorianj (approved):
This is somewhat under-specified on its own (could specify the exact models etc), however I think given the prior RFC that went out, and existing code, that figuring out the details in implementation should be fine.

rfcs/032-lineage-backend.md
NiravLangaliya: I agree, because we have started to define neo4j queries based up on pervious RFC and also lineage model available into amundsen-common. though these are not optimized queries and not completed but I was planning to put queries on this line. 

```
    def get_lineage(self, *,
                    id: str,
                    resource_type: ResourceType, direction: str, depth: int) -> Lineage:

        get_both_lineage_query = textwrap.dedent(u"""
        MATCH (down_parent:Table)<-[downstream_len:DOWNSTREAM*..{depth_key}]-(child:Table {{key: $query_key }})-[upstream_len:UPSTREAM*..{depth_key}]->(up_parent:Table)
        WITH
        child.key as child_key
        ,collect(distinct{{level:LENGTH(upstream_len),source:'hive',key:up_parent.key}}) AS upstream_entities
        ,collect(distinct{{level:LENGTH(downstream_len),source:'hive',key:down_parent.key}}) AS downstream_entities
        RETURN
        collect({{
        key:child_key,direction:"both",depth:1
        ,upstream_entities:upstream_entities
        ,downstream_entities:downstream_entities
        }}) AS lineageOutput
        """).format(depth_key=depth)

        get_upstream_lineage_query = textwrap.dedent(u"""
        MATCH (child:Table {{key: $query_key }})-[upstream_len:UPSTREAM*..{depth_key}]->(up_parent:Table)
        WITH
        child.key as child_key
        ,collect(distinct{{level:LENGTH(upstream_len),source:'hive',key:up_parent.key}}) AS upstream_entities
        RETURN
        collect({{
        key:child_key,direction:"upstream",depth:1
        ,upstream_entities:upstream_entities
        }}) AS lineageOutput
        """).format(depth_key=depth)

        get_downstream_lineage_query = textwrap.dedent(u"""
        MATCH (down_parent:Table)<-[downstream_len:DOWNSTREAM*..{depth_key}]-(child:Table {{key: $query_key }})
        WITH
        child.key as child_key
        ,collect(distinct{{level:LENGTH(downstream_len),source:'hive',key:down_parent.key}}) AS downstream_entities
        RETURN
        collect({{
        key:child_key,direction:"downstream",depth:1
        ,downstream_entities:downstream_entities
        }}) AS lineageOutput
        """).format(depth_key=depth)

        if direction == 'upstream':
            records = self._execute_cypher_query(statement=get_upstream_lineage_query,
                                                 param_dict={'query_key': id})
        elif direction == 'downstream':
            records = self._execute_cypher_query(statement=get_downstream_lineage_query,
                                                 param_dict={'query_key': id})
        else:
            records = self._execute_cypher_query(statement=get_both_lineage_query,
                                                 param_dict={'query_key': id})

        result = records.single()['lineageOutput'][0]
        return result
```

rfcs/032-lineage-backend.md
```diff
@@ -0,0 +1,97 @@
...
+The goal of this project is to have a lineage graph that is native to Amundsen so that users can find table and column lineage in one platform.
+
+## Motivation
+
+At the moment we are relying completely on any 3rd party lineage tool to show the table and column lineage in Amundsen. 
+Not everyone has access to such a 3rd party tool and want to have lineage information available natively in Amundsen.
+
+ref: https://github.com/amundsen-io/rfcs/blob/master/rfcs/025-lineage-stage-0.md
+
+## Guide-level Explanation (aka Product Details)
```
feng-tao, line 20: could you describe whether backend is only for graph or atlas or mysql? which one does it not plan to support?

rfcs/032-lineage-backend.md
```diff
@@ -0,0 +1,97 @@
...
+
+No new concepts/definitions are being introduced as a part of this RFC. 
+
+Databuilder already has the table lineage model, which creates an upstream/downstream relation to adding to the Neo4j graph.
```
feng-tao, line 24: do you plan to change the existing model?
If so, what will be the new model interface?

The current one doesn't take into account for job/application that generate the lineage in between, do you plan to add those?

feng-tao (changes requested):

rfcs/032-lineage-backend.md
dorianj: added

rfcs/032-lineage-backend.md
dorianj: clarified

dorianj (commented):
@feng-tao Added some clarifications to answer your questions, thanks for the feedback!

feng-tao (approved):

PR #33, opened 2021-04-01
Adds an RFC to change the lineage to graph visualisation

verdan:


dorianj:
Now that the mocks are attached, and given that this has been previewed and introduced to community at the community meeting without pushback. I think this is very much in line with community's expectation for how lineage would be displayed within Amundsen. Thus, I think would be appropriate to push to FCP.

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,101 @@
...
+- User will be able to Pane and Zoom, and also reset the whole graph.
+- Clicking any node will display the metadata details of that particular node in the sidebar.
+
+Important to note: We will not display the complete list of upstream/downstream tables. We will only display X number of tables by default based on their usage and will control the number by a config flag.
+
+See the Screenshots in [this folder](https://github.com/amundsen-io/rfcs/tree/98c240d29d8644a81b4eaf6a1481a126d08b5732/assets/033).
+
+## Reference-level Explanation (aka Technical Details)
+
+This does not change anything on the backend and uses the existing endpoints exposed in the v0 of lineage.
```
dorianj, line 61: I believe some new filtering APIs will be required, as will limiting of responses based on usage?

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,101 @@
...
+All the UI/UX explanations mentioned in the above section.
+
+## Drawbacks
+
+> Why should we _not_ do this?
+
+> Please consider:
+> Implementation cost, both in term of code size and complexity
```
dorianj, line 69: This is worth calling out. I believe the surface area of this implementation, if fully completed to spec, will have nearly as much interactivity and surface area as nearly the rest of Amundsen combined. I think it's worth it, however.

dorianj (approved):

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,103 @@
...
+> Integration of this feature with other existing and planned features
+> The impact on onboarding and learning about Amundsen
+> Cost of migrating existing Amundsen installations (is it a breaking change?)
+> If there are tradeoffs to choosing any path. Attempt to identify them here.
+
+## Alternatives
+
+> Why is this design the best in the space of possible designs?
+> What other designs have been considered and what is the rationale for not choosing them?
+> What is the impact of not doing this?
```
Golodhros, line 81:
```suggestion
- Keep the list based approach we currently have, showing only one level. This is a suboptimal way of representing this kind of data and transfers the burden to users, that will shortly grow tired of bouncing from page to page to have a good view of where data is coming.
```

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Prior art
+
+> Discuss prior art, both the good and the bad, in relation to this proposal. A few examples of what this can include are:
+
+> Does this feature exist in other data search applications and what experience have their community had?
+> For community proposals: Is this done by some other community and what were their experiences with it?
+> Papers: Are there any published papers or great posts that discuss this? If you have some relevant papers to refer to, this can serve as a more detailed theoretical background.
+
+> This section is intended to encourage you as an author to think about the lessons from other projects, provide readers of your RFC with a fuller picture. If there is no prior art, that is fine - your ideas are interesting to us whether they are brand new or if it is an adaptation from other projects.
```
Golodhros, line 91:
```suggestion
At Amundsen, we have the list based lineage that, as commented before, is a suboptimal way of showing this data.

We have seen this types of graphs on applications like DataHub, CompilerWorks and similar, so even if they don't   do a great job at it, it is clear that a graph is the right way of showing this information.
```

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Unresolved questions
+
+> What parts of the design do you expect to resolve through the RFC process before this gets merged?
+> What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
+> What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
```
Golodhros, line 97: I think we are missing a complete breakdown of the tasks to perform.
Maybe more details about the technology to use?

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,103 @@
...
+
+# Amundsen Lineage Stage 1 - Graph Visualization
+
+## Summary
+
+This is the next in the series of https://github.com/amundsen-io/rfcs/pull/25
+
+Implement a navigable canvas-based graph visualization for lineage. Lineage is traditionally represented this way, suboptimally, in the various existing tools like CompilerWorks, Castor, Apache Atlas, etc.
+
+This will replace the existing `Upstream` and `Downstream` tabs on the table detail page, and introduce a new CTA "Lineage" on the table detail page.
```
feng-tao, line 14: Given the lineage tab has just been merged and checked in, does it make sense to remove and replace with this without product verification ? 

cc @danwom

feng-tao: another way to ask is that whether it makes sense to have both until we found the graph UI is ready to replace the tab view .

verdan: The idea is to keep the list-based view in place for the time being, and the end state of this feature is to actually replace the list-based view with this graph visualization. 
I updated the summary section to make it explicit, and also added the configuration changed in the "Technical Details" section on how I think it will be during the development of this feature.

rfcs/033-lineage-graph.md
verdan: I've added the technical details in the Technical Details section to make it clear.

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,138 @@
...
+
+A map graph UI can more fully, transparently, and frictionlessly convey the lineage information. By co-locating data from multiple sources, Amundsen is positioned to create a better map graph UI experience vs. competitors.
+v0 of lineage is already available in the Amundsen project today in the form of a list.
+
+## Guide-level Explanation (aka Product Details)
+
+This feature will expose upstream and downstream tables and columns within the Table Details page, using a navigable graph visualization.
+
+- Lineage: Lineage is a term that describes the flow of data from one entity to another. While this term can broadly include everything from services, events, ETLs, and dashboards, we will focus on table-to-table and column-to-column data lineage in this RFC.
+- Upstream: Upstream is a relative term that describes data sources from which we inherit. Data flows from upstream to downstream.
```
mgorsk1, line 28: maybe this is also opportunity to think about word choice here? upstream/downstream are terms with which you need to spend some time to feel familiar with, they seem very `technical`. for example - in Atlas, the word instead of `downstream` there is `impact` - more closely aligned with what lineage is all about.

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,138 @@
...
+- RFC PR: [amundsen-io/rfcs#33](https://github.com/amundsen-io/rfcs/pull/33)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
+
+# Amundsen Lineage Stage 1 - Graph Visualization
+
+## Summary
+
+This is the next in the series of https://github.com/amundsen-io/rfcs/pull/25
+
+Implement a navigable canvas-based graph visualization for lineage. Lineage is traditionally represented this way, suboptimally, in the various existing tools like CompilerWorks, Castor, Apache Atlas, etc.
```
mgorsk1, line 12: is there any research explaining shortcomings of existing tools? would be good to understand on which of them we are trying to improve. for one, I think Apache Atlas lineage view is actually pretty good

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,138 @@
...
+- a new CTA/Button will appear on the top right of the table detail page.
+- a new icon in front of the column name (to support column-level lineage).
+
+Clicking on any of the above will navigate the user to a new page displaying lineage graph along with the filters to select the `level` and `direction` for the graph.
+
+The sidebar will have the filters available and will be used for other interactions like displaying a complete list of nodes, and/or view the metadata of a selected node.
+
+Interaction Details:
+
+- By default, we will display 1 level (or depth) of the lineage. Users will be able to control the level by clicking Expand/Collapse via each node. Or simply use the filter to change the number of levels shown.
```
mgorsk1, line 51: - default level could also be configurable through config-custom.js
- `filter` to change the number of levels could be replaced by `slider` for better UX

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,138 @@
...
+- User will be able to Pane and Zoom, and also reset the whole graph.
+- Clicking any node will display the metadata details of that particular node in the sidebar.
```
mgorsk1, line 53: is it going to require extension for lineage model or separate api call to metdata to collect the lineage?

rfcs/033-lineage-graph.md
```diff
@@ -0,0 +1,138 @@
...
+
+Important to note: We will not display the complete list of upstream/downstream tables. We will only display X number of tables by default based on their usage and will control the number by a config flag.
```
mgorsk1, line 55: just a side-question: would there be a tie-breaker if all tables have same popularity?

rfcs/033-lineage-graph.md
verdan: I'd keep this upstream/downstream for now.

rfcs/033-lineage-graph.md
verdan: Atlas lineage is pretty extensive and is good. I agree. And we can improve on these things based on the user feedback. Controlling the depth and clustering the nodes is the advantage over the other tools i.e., showing top X nodes, etc.

rfcs/033-lineage-graph.md
verdan: default level should be a config variable indeed.

rfcs/033-lineage-graph.md
verdan: This will be a call to metadata endpoint to maintain the performance.

rfcs/033-lineage-graph.md
verdan: I don't have an answer for this atm. I think if usage is same, we can think of sorting alphabetically then.

mgorsk1 (approved):

PR #34, opened 2021-04-16
Support MySQL as the metadata backend in metadata service

crazy-2020:


rfcs/000-mysql-support-in-metadata-service.md
```diff
@@ -0,0 +1,56 @@
...
+The support includes the database initialization and a new proxy for MySQL in metadata service.
+
+## Motivation
+
+Metadata can be published into mysql from Amundsen databuilder now and this RFC is going to make metadata retrieval API work
+for MySQL in metadata service.  
+
+## Guide-level Explanation (aka Product Details)
+
+- The database initialization will be supported by CLI. The command will init/upgrade table schemas according to
```
feng-tao, line 19: so you will add a new CLI to do db migration (e.g similar to airflow db init?)? when the wsgi proxy starts, how the proxy checks whether the db schema is inited successfully?

feng-tao (approved):

feng-tao:
lgtm

rfcs/000-mysql-support-in-metadata-service.md
crazy-2020: yeah, it might be a flask custom command, which may be managed by some entrypoint script for docker use. For the verification of schema init, I haven't thought about it. If there will a check, it should be about checking the latest version number with alembic script head. I'll think about that. Thanks.

feng-tao: sg

PR #36, opened 2021-05-05
ML Features Discovery in Amundsen

allisonsuarez:
RFC outlining the work needed for an MVP version of [ML project discovery](https://github.com/amundsen-io/rfcs/pull/8).


rfcs/000-ml-feature-discovery.md
```diff
@@ -0,0 +1,168 @@
+- Feature Name: ml_feature_discovery
+- Start Date: 2021-05-05
+- RFC PR: [amundsen-io/rfcs#0000](https://github.com/amundsen-io/rfcs/pull/0000) (after opening the RFC PR, update this with a link to it and update the file name)
```
Golodhros, line 3:
```suggestion
- RFC PR: [amundsen-io/rfcs#36](https://github.com/amundsen-io/rfcs/pull/36)
```

Golodhros: did this change?

allisonsuarez: I clicked commit change, this is weird I'll try again

allisonsuarez: done!

allisonsuarez: okay now done for real hahaha

rfcs/036-ml-feature-discovery.md
```diff
@@ -0,0 +1,195 @@
...
+Integration of ML projects in Amundsen has been heavily requested by the community and has already been implemented by other similar projects. This RFC outlines what will be the first step towards indexing ML projects in Amundsen since feature engineering is one of the most time consuming processes for ML engineers. Providing them with the ability to easily discover features that they can reuse for their models will significatly increase their productivity.
+
+## Guide-level Explanation (aka Product Details)
+
+### New Metadata Definitions
+#### Semantic Metadata
+Metadata related to the feature itself.
+- Key: what makes a feature uniquely identifiable is the group, name and version. We don’t use db/source because the same feature can exists in different places (like hive and dynamo)
+    - ex: test_feature_group_name/test_feature_name/1.2.0
+- Availability/sources: list of places where feature is available
```
feng-tao, line 23: could you describe the scenario on why the same feature could be available in different places(hive, dynamo, etc)? are they generated by different pipelines(streaming,batch) to different sinks with the same IDL schema?

rfcs/000-ml-feature-discovery.md
```diff
@@ -0,0 +1,183 @@
...
+- Name: the name of the feature
+- Version: the version of the feature
+- Data type
+- Description: optional, a description of the feature
+- Owners: individual and team emails of owners
+- Feature group: a way of grouping features, can be defined differently in each company, in this case we assume this will be the name of the pipeline that produced the features
+- Entity: a way of grouping features semantically by using nouns
+- Owner tags: programmatically generated and not editable tags, useful to group features base don the specific engineering team using them
+- User tags: normal amundsen tags
+- Contains PII: weather a feature contains PII or not, will be a badge in the badges list
```
sewardgw: Most of the features I've used in the past are some quantified value they are aggregated (e.g. avg daily balance, mean # of rides, log(daily sales), etc.) or they are transformed (e.g bins, categorical variables, etc.) where any form of PII data is removed as part of the calculation. Could you provide examples for how Contains PII would be used here?

rfcs/036-ml-feature-discovery.md
```diff
@@ -0,0 +1,195 @@
...
+- Name: the name of the feature
+- Version: the version of the feature
+- Data type
```
feng-tao, line 26: could you add some examples of what data type for feature you have in mind?

rfcs/036-ml-feature-discovery.md
```diff
@@ -0,0 +1,195 @@
...
+- Description: optional, a description of the feature
+- Owners: individual and team emails of owners
+- Feature group: a way of grouping features, can be defined differently in each company, in this case we assume this will be the name of the pipeline that produced the features
+- Entity: a way of grouping features semantically by using nouns
+- Owner tags: programmatically generated and not editable tags, useful to group features base don the specific engineering team using them
```
feng-tao, line 31: why not badge?

rfcs/036-ml-feature-discovery.md
```diff
@@ -0,0 +1,195 @@
...
+- User tags: normal amundsen tags
+- Contains PII: weather a feature contains PII or not, will be a badge in the badges list
+- Date range: what time frame is the data in the feature for
+- Partition column: optional column associated to the feature to indicate partition
```
feng-tao, line 35: badge?

rfcs/036-ml-feature-discovery.md
```diff
@@ -0,0 +1,195 @@
...
+
+#### Operational Metadata
+Metadata to facilitate the use of the feature by other engineers.
+- Code snippet/SQL: snippet of code that generated the feature and external URL to more details
+- Last updated: timestamp for when the feature was most recently updated
+- Usage: number of accesses to the feature, usage in models out of scope for MVP since we are not indexing models yet
+- Upstream resources: upstream lineage in the form of tables, could be expanded to anything else in the future (column, service, pipeline)
```
feng-tao, line 42: a few lineages (producer to ML feature, consumer of ML feature, source table to feature table lineage, etc)

rfcs/000-ml-feature-discovery.md
```diff
@@ -0,0 +1,183 @@
...
+## Alternatives
+
+- Create a generic entity that can be used as a feature.
+
+## Prior art
+
+- [Datahub ML models](https://github.com/linkedin/datahub/tree/master/docs/rfc/active/1812-ml_models)
+- [Feast](https://github.com/feast-dev/feast)
+
+## Unresolved questions
```
sewardgw: It looks like this RFC is able to support materialized features well - there are attributes for where the feature is persisted , the query that uses it, etc. There are are other features that are calculated on the fly, just before model execution (e.g. with scikit learn / TF pipelines, custom transformations, etc.) which may be transient and I am curious how how well you see these features being supported. 

While I think many of these on-the-fly features may be unique to the model that they are associated this could also be an area where feature discovery helps to understand how these "last-mile" features are created.

sewardgw (commented):
👏🏻  Love this! Left some comments / questions in line

rfcs/036-ml-feature-discovery.md
allisonsuarez: Knowl said badges are not scalable design wise because of where they are located, these sit in the middle of badge and tag

rfcs/036-ml-feature-discovery.md
allisonsuarez: some examples for us are bigint, string, map but since it is represented as a string it can be anything

rfcs/036-ml-feature-discovery.md
allisonsuarez: if you read below you'll see that we are leveraging badge on the model to indicate this

rfcs/036-ml-feature-discovery.md
allisonsuarez: right now we are just doing upstream tables since indexing models is not in scope for MVP, definitely in plans for future work

rfcs/000-ml-feature-discovery.md
```diff
@@ -0,0 +1,183 @@
...
+- In the future this feature will hopefully replace all use of old Tag badges.
+
+### New Metadata Definitions
+#### Semantic Metadata
+Metadata related to the feature itself.
+- Key: what makes a feature uniquely identifiable is the group, name and version. We don’t use db/source because the same feature can exists in different places (like hive and dynamo)
+    - ex: test_feature_group_name/test_feature_name/1.2.0
+- Availability/sources: list of places where feature is available
+- Name: the name of the feature
+- Version: the version of the feature
```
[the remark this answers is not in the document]
allisonsuarez: I think in the future we can consider something like entity history for all indexed entities, but for now version is just meant to allow differentiation for features (we can index the same feature with different versions as 2 different features). I believe what you are suggesting is a necessary next step for Amundsen but this will require more careful consideration separate from this work.

rfcs/000-ml-feature-discovery.md
```diff
@@ -0,0 +1,183 @@
...
+- Feature node:
+    - Key
+    - Name
+    - Version
+    - Feature group
+    - Entity
+    - Data type
+    - Last updated
+#### Relationships
+Here we show the relationships between feature and other nodes, as well as the metadata that this represents:
```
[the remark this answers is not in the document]
allisonsuarez: We are currently doing this by having an upstream tab that relates to tables rather than specific columns, in the future we can expand this to be both or even integrate upstream services rather than datasets.

rfcs/000-ml-feature-discovery.md
```diff
@@ -0,0 +1,183 @@
...
+```
+class Feature:
+   key: str  # ex: test_feature_group_name/test_feature_name/1.2.0
+   name: str
+   version: str  # ex: "1.2.0"
+   status: str
+   feature_group: str
+   entity: Optional[List[str]]
+   data_type: Optional[str]
+   availability: List[str]
```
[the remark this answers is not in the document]
allisonsuarez: I think this is already included in upstream resources. The issue with making it the same in availability is that we might lost important metadata if the resource is not listed in Amundsen, for example we don't index dynamo stuff yet but we want to tell ML engs that this feature is in fact available there.

rfcs/000-ml-feature-discovery.md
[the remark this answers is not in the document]
feng-tao: Versioning is sth which requires more producer/ user thought. E.g what scenarios, how user consume version, why users prefer seeing previous version , etc, how to clean up. I think it is out of scope of this PR.

rfcs/000-ml-feature-discovery.md
```diff
@@ -0,0 +1,183 @@
...
+   name: str
+   version: str  # ex: "1.2.0"
+   availability: List[str]
+   entity: Optional[List[str]]
+   description: Optional[str]
+```
+### Amundsen Metadata
+The features API methods will basically be identical to that for tables, but the response will follow the outline of the common schema specified above.
+#### Requests
+- GET feature details
```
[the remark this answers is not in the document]
allisonsuarez: each of them would have their own page right now given that we don't have a comprehensive design for history of a dataset, feature, dashboard, etc.

rfcs/000-ml-feature-discovery.md
allisonsuarez: I think this varies depending on how you source your data, for us Amundsen simply indexes features from a different service at Lyft, so it is the responsibility of the feature creator to specify this.

rfcs/036-ml-feature-discovery.md
amalakar: We do have cases where we generate offline features using hive/presto etc. The same feature then also gets loaded into dynamodb.

So the same feature end up in multiple stores, and depending on the context ML model execution may choose the store (low latency use dynamodb, high latency is acceptable but high throughput is needed use hive) that is more appropriate. For example ride_count_30_days as a feature may get used by an online model for fraud detection which has low latency requirement and uses dynamodb. The exact same feature also gets used in batch model predictions for weekly coupon distribution which uses hive (or presto) for feature retrieval.

rfcs/000-ml-feature-discovery.md
sewardgw: I agree that the creator should have the responsibility to specify if something is PII or not. In the same vein, why not just have the feature creator add a badge directly instead of having an attribute dedicated to PII?

feng-tao:
cc @dwarszawski  who created https://github.com/amundsen-io/rfcs/pull/8 , I wonder we should just close the initial RFC for this one.

rfcs/036-ml-feature-discovery.md
kbaekgit: Yes, the purpose of Feature Owner Tags is to serve as a first step towards providing a 'semantic grouping' of features vs. the Feature Group which is based on the pipeline that produced the features. 
We found out through user research that they want features to be organized in semantic grouping with clear hierarchy, but until we figure out the mater plan, this will allow feature owners to place tags that are more useful to consumers e.g. teams who will use the features.
We decided not to use Badges because there could be so many of these team names and other semantic grouping, and currently we limit the use of badges to things like Resource Status (Beta/GA..) or Column Type (Partition Column etc).

rfcs/036-ml-feature-discovery.md
kbaekgit: Yes PII will be a badge!

rfcs/000-ml-feature-discovery.md
allisonsuarez: if you look at the common schema specified in the RFC there is no pii field, rather a badges list, so I completely agree.

rfcs/000-ml-feature-discovery.md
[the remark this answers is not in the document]
sewardgw: After thinking through this a bit more, I really like the definition of feature groups. It really resonates with how I've worked on teams in the past that associate features to the pipeline that creates them, the entity they represent, etc. From a UX perspective, I think displaying features at the feature group instead of having one page for each feature could be better (if you _had_ to map this to today's world feature group = table, feature = column).

The thought here is that the highest impact discoverability will occur immediately after a user finds a feature that is relevant to them. The feature group would inherently contain similar features and help augment the discoverability of features.

Thoughts?

rfcs/000-ml-feature-discovery.md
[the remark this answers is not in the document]
sewardgw: I see your point, there could be a lot of information loss here. 

Given that this field just a string and the upstream information is available as you mentioned, I'm wondering if this needs to be a first class object or this is an area where the implementer could use prog desc to provide this info. I'd be curious how much others in the community see this being used.

feng-tao (approved):

sewardgw (approved):
👍🏻

dwarszawski:
> cc @dwarszawski who created #8 , I wonder we should just close the initial RFC for this one.

Fine by me.

rfcs/000-ml-feature-discovery.md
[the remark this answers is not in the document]
allisonsuarez: Since it is a list just having it be [] and not provide the information wouldn't really make a big difference, we could make it a prog description too. This is the use case we had internally for it, if it doesn't come up for anyone else then maybe we can remove it: https://github.com/amundsen-io/rfcs/pull/36#discussion_r627658672 what do you think?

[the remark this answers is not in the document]
sewardgw: Sounds good, lets see if anyone else in the community can chime in to share an external perspective.

rfcs/000-ml-feature-discovery.md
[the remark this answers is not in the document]
santok: @sewardgw, you are absolutely right about feature groups when conceptualizing them as feature group = table and feature = column. A large # of features have been derived from Hive tables historically. 

However, we are increasingly witnessing online features derived from streaming sources. There this concept of feature groups break down. Features may not be generated via a single pipeline and may not be contained inside a single encapsulating container such as a table. Displaying such features has to be done without a concept of a feature group. 

Therefore, feature group is an optional metadata. And we decided to display features as independent entity. However, if one wishes to look at all features corresponding to a given feature's feature group, we will enable that. Let me know if this makes sense.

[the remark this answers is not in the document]
sewardgw: Makes sense! Love the thought around having options for iterations on this feature to continue the buildout if needed.

PR #2, opened 2020-08-18
Adds Feature and Deprecation templates

Golodhros:
Adds templates for RFCs

Signed-off-by: Marcos Iglesias <golodhros@gmail.com>

PR #37, opened 2021-05-13
Index queries and query compositions

sewardgw:
Signed-off-by: Grant Seward <grant@stemma.ai>

This RFC cleans up, refines and, in some cases, expands on the concepts put forth in https://github.com/amundsen-io/rfcs/pull/27.


dorianj:
Strongly in favor of this proposal. When I put forward the previous proposition, it was rather under-specified in how it would actually be used, and it forced a particular implementation style.

This solution is more elegant because it works great with multiple integration types.

In particular, this allows for the integrator quite a bit of flexibility in terms of how much data they want to put in. If they have a very large amount of query logs stored in a different system, they can materialize just a small aggregated view into Amundsen of top queries. OTOH, if their volume is lower, this system will accommodate simply loading everything and the read path will sort by usage.

rfcs/037-index-query-composition.md
```diff
@@ -0,0 +1,115 @@
...
+
+![Query Assets](../assets/037/query-assets.png)
+
+1. **Query**: these contain the full SQL for a query as well as relationships to Tables, Users, Query Executions, Joins and Wheres
+2. **Query Execution**: this represents the total number of times a query was executed within a time window. By having query executions separated from queries Amundsen can keep one copy of the query in the database and sum the total number of executions at any point in time to get the total execution count. Query Executions can be added or dropped to add or remove relevance to a Query, Join or Where
+3. **Join**: this is a component of a query, a single join clause (e.g. `table_1 a join table_2 b on a.field = b.field`). Joins are connected to the columns they are joined on as well as to queries. By connecting to queries, Amundsen can rank order the most frequent joins by accessing the query executions.
+4. **Where**: this is a component of a query, a single where clause (e.g. `where field_a is not null`). Wheres are connected to columns as well as to queries. By connecting to queries, Amundsen can rank order the most frequent wheres by accessing the query executions.
```
feng-tao, line 32: I am not sure actually, for join relationship it makes SENSE. For where , it seems to be a bit specific use case. Could you talk about this use case? And there are many different sql paradigm as well(e.g ctas), why only where clause?

rfcs/037-index-query-composition.md
```diff
@@ -0,0 +1,115 @@
...
+        - left_column: Column Metadata for the left column in a join
+        - right_column: Column Metadata for the right column in a join
+        - join_type: a string representing the join type, could be `inner join`, `left outer join`, etc
+        - join_operator: a string representing the operator, could be `=`, `>`, `<=`, etc.
+        - join_sql: the sql code for this join (e.g. `table join table2 on table.a = table2.b`)
+        - query: The query model defined above
+    - Relations:
+        - Column >> Join
+        - Query >> Join
+4. Where
```
feng-tao, line 82: again, I am not sure about whether this is useful. Could you talk about some real-life example from an user standpoint how user to use this where clause info?

rfcs/037-index-query-composition.md
```diff
@@ -0,0 +1,115 @@
...
+Four new models will be added to support the data assets listed above:
+
+1. Query
+    - Attributes:
+        - sql: the SQL code for the query
+        - tables: a list of Table Metadata representing each table used in the query
+        - clean_sql: an optional SQL string that the implementer has curated. For example, if an Airflow DAG injects the date into a SQL query every day, the implementer may strip that date and pass the updated SQL into `clean_sql` in order to allow Amundsen to identify that this is not a unique query.
+        - user: an optional user that executed the query (should be Amundsen user ID)
+    - Relations:
+        - Table > Query
```
feng-tao, line 59: currently the dashboard model has query model already. How do we de-duplicate on table-query and dashboard-query?

also how do you plan to figure the table<->query relationship (and what is considered to have query and table relationship)? is it through sql parsing?

rfcs/037-index-query-composition.md
```diff
@@ -0,0 +1,115 @@
...
+Parsing query logs can be a large effort, therefore this approach will be designed in a way such that implementers will be able to send a raw query into Amundsen without the other components (executions, joins and wheres) but will loose the ability for the features built on top of those data assets such as ordering by most common query, showing the most common filters, etc. However, Amundsen is not intended to hold a copy of each query each time it is executed and implementers should take some effort to dedupe queries.
+
+## Alternatives
+
+- Do not index queries and query components
+- Only index queries, not components of queries
+
+## Prior art
+
+[PR 27](https://github.com/amundsen-io/rfcs/pull/27) had been opened to add queries as first class objects. This PR did a good job to raise awareness of the need for having queries as first-class objects but this PR expands on that in a few areas:
```
feng-tao, line 102: are you talking about making query as entity like table? what is the benefit compared to showing the top used query in table tab / dashboard tab?

feng-tao: I haven't seen any other solutions has been put query as 1st class entity. The user benefit is unclear compared to putting query under table<->dashboard which are more user-phasing( to me query node is similar to column node that belong to table/dashboard. )

feng-tao (changes requested):

rfcs/037-index-query-composition.md
feng-tao, line 102: I put quite some concerns over https://github.com/amundsen-io/rfcs/pull/27 . Please take a look.

rfcs/037-index-query-composition.md
dorianj: The use case we've seen is an analyst or data engineer who is trying to figure out if there are any common filter conditions. Usually that's to see if there are any caveats with the data (e.g. you basically always need to filter out NULLs from the column, or you need to exclude this status to make sure to only have prod data). Sometimes this is used to identify places to add new aggregate tables (e.g. if everyone is using this filter, might make sense to create a new derived table).

When you say CTAS, do you mean `create table as select?` If so, I think the way we'd like to ingest that is as lineage, which is out of scope for here

rfcs/037-index-query-composition.md
dorianj: added in other thread https://github.com/amundsen-io/rfcs/pull/37#discussion_r632748662

rfcs/037-index-query-composition.md
dorianj: > currently the dashboard model has query model already. How do we de-duplicate on table-query and dashboard-query?

good question, I'll let @sewardgw respond to that


> also how do you plan to figure the table<->query relationship (and what is considered to have query and table relationship)? is it through sql parsing?

Sort of. Stemma has full SQL parsing capabilities, but it's proprietary. Our hope is to provide the plumbing such that OSS implementors could easily sub in their own parsing capabilities. We've seen that many companies possess this capability already to varying degrees of veracity: some have simple regex based scripts that can extract basic info about the query (e.g. joins); we've seen other companies with ANTLR-based parsers with which they extract more fine-grained details.

rfcs/037-index-query-composition.md
dorianj: Sorry, I used unclear language and gave the wrong impression.

The intention is to make queries a top level concept at the data level, but not in the UI. Queries cannot be conceptually under tables (like a column can), because a single query touches multiple tables. However, we certainly would display them that way (on the table detail page, as you mentioned)

We will revise the RFC to make this clearer.

dorianj: let me pull them in so they're in one place 

> typically approach is to store the query log through a data warehouse table (delta, hive etc) and it could fetch the usage log and persist the relevant information into the metadata store.
> I wonder how. the usage query log ingestion is going to look like? are you directly consume from the source and directly persist into metastore store?

This is supported in the updated proposal. Some systems have it builtin, but most installs we've seen use a streaming service to dump the log data into an object store (e.g. Kinesis into S3), then they do batch aggregations into Amundsen. But overall it's been pretty disparate, and I don't feel super confident that we'll be able to build turn-key integrations that will work great for all installs. So we'd like to make all the building blocks available, and leave the actual aggregation and processing up to the implementor so they can fit it into their infrastructure.


> have we done any tests on how long it takes to ingest user log sql. Typically production usage log is 10s of millions of sql per day. Are we planning to surface all the queries in the table page?

The table detail page will only have the top 2-10 queries, configurable. I don't even think it needs to be paginated, a simple leaderboard is sufficient for most of the use cases we've seen.

> if not during ingest, when will the sql parsing happen? are we planning to expose the usagelog for lineage consumption through API?

Can you clarify your question about usage log for lineage?


> could you clarify the audit logging piece? also will the usage log ingestion refreshed everyday?

We're de-pri-ing the audit log use case here, since we're no longer storing every single invocation.


> there is already a https://github.com/amundsen-io/amundsendatabuilder/blob/master/databuilder/models/dashboard/dashboard_query.py, we should at least make a base query class and make two subclasses (dashboard, table)

👍 


> I think having a table query relationship makes sense, however, I am not sure about using amundsen as the source of truth for all the queries which could easily store many non-sense queries(e.g typo etc).

> SLA of query log emit from computation framework to this store is NA (typical is to hive table)
> Non GC of the query nodes, are you going to retain PII data(query has all kinds PII) for the full deployment?
> I think we could have the query model, but if I am going to use it, I will only surface the most relevant / popular queries to the table (top 5, or top X)

Yep, this should all be easy to support with these models. Basically, implementors can GC the query agg nodes after 7-30 days (whatever they want) in the graph, and then all of the orphan nodes can be GC'd.

And because the ingest layer is deciding what to surface (and pre-aggregating), the total amount of data is bound.

rfcs/037-index-query-composition.md
sewardgw: @feng-tao could you please elaborate on `I am not sure actually, for join relationship it makes SENSE`. I'm not sure I completely follow the intent behind the question.

In regards to the where clause, the response from @dorianj is spot on but I will elaborate on the point about it being a specific use-case. The structure of this where-clause is actually set up such that it can capture relationships between any number of tables and columns in order to handle both complex as well as simple scenarios. This setup supports the following (and more):

- Simple:
```sql
where table1.field = 'value'
```
- Multiple columns
```sql
where table1.field > table2.another_value * 10
```
- Sub query
```sql
where table1.date = (select max(dt) from date_tbl)
```

Let me know if this helps.

rfcs/037-index-query-composition.md
sewardgw: At a high level what we're trying to do is to help users discover _how to use their data_ at the same time that they discover the location and meaning of their data. 

As to why create these as first-class objects in the database, there are some questions that are difficult to answer without it:

- What are the most common filters used _now_, this is especially important as data evolves over time (supported by `(Where) -> (Query) -> (Query Execution)` relationship)


In addition, there is the added benefit of referencing a single query object, a single join object, etc. instead of replicating that for each table / dashboard that it is associated to.

rfcs/037-index-query-composition.md
feng-tao: For SQL parsing, I wouldn't recommend `ANTLR-based` parser. Now we delete those extractors, but in the early days, we have presto Antlar parser which causes so many issues in production.

sewardgw: Re: `how do you plan to figure the table<->query relationship ` We are not dictating _how_ to do this. This is definitely a complicated task but there are members of the community doing this already. We are providing the pipes to enable them to structure, access and display the information within Amundsen.

sewardgw: Re: `How do we de-duplicate on table-query and dashboard-query`. Thanks for bringing this up, this is a great point. As an FYI There is also a "query" object being created as part of the ML Features PR (e.g. query used to create features.) so this question is especially relevant.

 There are some unique attributes of the dashboard query model that we do not want to lose (e.g. `dashboard_group_id`, `dashboard_id`). I think the best options are:
- One common query model that all "query" use-cases use, or
- One common Query Model that the others inherit from but add their own specific attributes (keeping the same node label / relationship labels)

My preference would be toward #2 but would love feedback before adding this additional context to the PR.

sewardgw: Here is the other PR: https://github.com/amundsen-io/amundsen/pull/1059/files

Tagging @allisonsuarez here as well

dorianj: > For SQL parsing, I wouldn't recommend ANTLR-based parser. Now we delete those extractors, but in the early days, we have presto Antlar parser extractor which causes so many issues in production.

Yeah, we're not planning to build this into Amundsen -- just echo'ing what I've seen companies do in the wild

allisonsuarez: @dorianj @feng-tao do you think the Query schema here is generic enough to be used for other resources? https://github.com/amundsen-io/amundsen/pull/1059/files#diff-4a840835d48ad91e06402d8bdcb555dcd2ed2c40457c59713ebb5c83fda64bdaR16

sewardgw: I will add my 2cents in that I _do_ think the implementation from @allisonsuarez  it is generic enough (there aren't a lot of fields so the minimal implementation lends itself to be flexible). I would propose putting it in a different location though, specifically since this RFC is building out custom components of queries I would propose a query-specific file.

allisonsuarez: it is worth noting that this is basically the same model we already use for Dashboards. I will move it to its own file so we can unblock the ML features work.

rfcs/037-index-query-composition.md
sewardgw: @feng-tao just wanted to bump this to see if it answers your questions or if you're able to add some additional context.

feng-tao: I think adding a query model to show it in table optional makes sense. I am just not convinced showing `where` relationship is useful for end-user. But it is not a blocker from my side.

feng-tao: or at the very least that needs to make it more generic like `query->transform->table` etc

feng-tao (approved):
assume the RFC is updated that it is not purposing query as another first class  resource entity like table

sewardgw:
Just to confirm, when you say "first-class resource entity like table", do you mean something that would be a searchable item in the UI, have it's own dedicated page, etc? If that is the case, then no, this RFC would not make it a first-class entity in that sense; this RFC only lays the structure for the data models and relationships for queries and their components.

dorianj (approved):

PR #38, opened 2021-05-22
Unify shared dependencies

mgorsk1:
Signed-off-by: at91mm <mariusz.gorski@ing.com>

### DIFF
--- a/rfcs/038-unify-shared-dependencies.md
+++ b/rfcs/038-unify-shared-dependencies.md
@@ -106,7 +106,7 @@ We test this with `make test` on each subproject + building Amundsen locally and
 
 ## Unresolved questions
 
-n/a
+- Should we scope all subdirectories with that change or would we be ok excluding some (maybe common and databuilder)
 
 ## Future possibilities
 
--- a/rfcs/038-unify-shared-dependencies.md
+++ b/rfcs/038-unify-shared-dependencies.md
@@ -74,6 +74,7 @@ pip3 install ".[all]"
 10. In CI/CD apart from `make install` do also `pip3 uninstall amundsen-common && pip3 install ../common` (apart from common subdirectory which would not re-install ../common)
 
 ### Testing
+
 We test this with `make test` on each subproject + building Amundsen locally and testing functionalities manually relying on test data
 
 ## How We Communicate This
@@ -106,8 +107,8 @@ We test this with `make test` on each subproject + building Amundsen locally and
 
 ## Unresolved questions
 
-- Should we scope all subdirectories with that change or would we be ok excluding some (maybe common and databuilder)
+- Should we scope all subdirectories with that change or would we be ok excluding some and focusing on microservices only (for `requirements-common.txt`)
 
 ## Future possibilities
 
-n/a
+- Easier upgrade possibilities (for example bumping to Flask 2+)
--- a/rfcs/038-unify-shared-dependencies.md
+++ b/rfcs/038-unify-shared-dependencies.md
@@ -49,6 +49,7 @@ Multiple times you might find the term `subdirectory`, which refers to following
 
 Steps:
 
+0. Undertake new approach where all dependencies land in `install_requires` or `extras_require` through reading in appropriate requirements* files (except databuilder due to it's complexity)
 1. Split `requirements.txt` for each subdirectory into `requirements.txt` and `requirements-dev.txt` (there is no need to install flake8 when doing for example `pip install amundsen-metadata` which is currently the case)
 2. Identify libraries that are shared across all (or majority - see `Drawbacks`) of subprojects
 3. Create `requiremnts-common.txt` file in root folder of amundsen repo containing pinned dependencies for shared libraries
--- a/rfcs/038-unify-shared-dependencies.md
+++ b/rfcs/038-unify-shared-dependencies.md
@@ -1,0 +1,111 @@
+- Feature Name: Unify shared dependencies
+- Start Date: 2021-05-22
+- RFC PR: [amundsen-io/rfcs#38](https://github.com/amundsen-io/rfcs/pull/38) (after opening the RFC PR, update this with a link to it and update the file name)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000) (leave this empty for now)
+
+# Unify shared dependencies
+
+## Summary
+
+Major motivation behind migrating to monorepo was to solve `Dependency Management` issue. However, it still remains unsolved.
+
+This RFC serves as a platform to figure out the best approach for unifying shared dependencies and contains preliminary proposal for how that could be solved.
+
+## Motivation
+
+Following original RFC:
+
+```
+Dependency Management: Python packages are pretty much the same for each proxy, which makes it hard to sync across all the above repositories. As a result most of the time, each proxy is running its own version of the dependency.
+A few examples:
+
+| Package             | amundsenmetadata   | amundsenfrontend  |
+| --------------      |:------------------:| -----------------:|
+| amundsen-common     | >=0.8.1            | ==0.6.0           |
+| flake8              | ==3.5.0            | ==3.8.4           |
+| flake8-tidy-imports | ==1.1.0            | ==4.2.1           |
+
+and many more... This change will put all the packages in just one place, making practically to keep dependencies similar accross verisons, simply because it will be possible to open a single PR that updates multiple packages. Practically speaking this will improve the status quo slightly.
+```
+
+## Guide-level Explanation (aka Product Details)
+
+n/a
+
+## UI/UX-level Explanation
+
+n/a
+
+## Reference-level Explanation (aka Technical Details)
+
+Following steps are proposed to achieve goal of this RFC. 
+
+Multiple times you might find the term `subdirectory`, which refers to following folders:
+- common
+- databuilder
+- frontend
+- metadata
+- search
+
+Steps:
+
+1. Split `requirements.txt` for each subdirectory into `requirements.txt` and `requirements-dev.txt` (there is no need to install flake8 when doing for example `pip install amundsen-metadata` which is currently the case)
+2. Identify libraries that are shared across all (or majority - see `Drawbacks`) of subprojects
+3. Create `requiremnts-common.txt` file in root folder of amundsen repo containing pinned dependencies for shared libraries
+4. Remove libraries listed in `requirements-common.txt` from `requirements.txt` of each subdirectory
+5. Create `requirements-dev.txt` file in root folder of amundsen repo containing pinned dependencies for shared dev libraries (mypy, flake8, etc.)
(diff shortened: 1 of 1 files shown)


rfcs/038-unify-shared-dependencies.md
```diff
@@ -0,0 +1,112 @@
...
+- As we don't have 100% test coverage it's difficult to asses complete impact of evening out dependencies so we might enter period of ironing out potential issues
+- Using symlinks might be tricky for users using Windows
+- There are dependencies that are shared by some but not all libraries. For those we have two choices:
+    - Include them in `requirements-common.txt` (preferred by the author) - this will make some packages contain deps that are not used
+    - Populate `requirements-common.txt` only with deps used by all packages - this will still leave a lot of room for inconsistencies
+- It might be difficult to keep the dependencies balanced especially if some custom entry in one of subdirs `requirements.txt` requires different version of lib from `requirements-common.txt`
+
+## Alternatives
+
+- Using pip functionality `constraint` but I've identified this as less intuitive for end users as:
```
dorianj, line 95: I'm not familiar enough with python packaging to have an answer here, but it is a bit of a shame we can't use constraint files -- it feels like the "correct" solution. But I do agree it's not intuitive, and even if we published build scripts everywhere, many people wouldn't use them because there is an expectation you can just `pip install -r requirements` and be off to the races

If instead of including requirements.txt, we only used `setup.py` deps + constraint file, i wonder if it's possible to automatically bring in the constraint file in setup.py or setup.cfg ?

rfcs/038-unify-shared-dependencies.md
```diff
@@ -0,0 +1,112 @@
...
+
+Steps:
+
+1. Split `requirements.txt` for each subdirectory into `requirements.txt` and `requirements-dev.txt` (there is no need to install flake8 when doing for example `pip install amundsen-metadata` which is currently the case)
+2. Identify libraries that are shared across all (or majority - see `Drawbacks`) of subprojects
+3. Create `requiremnts-common.txt` file in root folder of amundsen repo containing pinned dependencies for shared libraries
+4. Remove libraries listed in `requirements-common.txt` from `requirements.txt` of each subdirectory
+5. Create `requirements-dev.txt` file in root folder of amundsen repo containing pinned dependencies for shared dev libraries (mypy, flake8, etc.)
+6. Remove libraries listed in `requirements-dev.txt` from `requirements.txt` of each subdirectory
+7. Create symbolic link to `requirements-common.txt` and `requirements-dev.txt` in each subdirectory
```
dorianj, line 58: Last time I checked, symlinks didn't work on Windows :/ -- do you know if that's been changed?

rfcs/038-unify-shared-dependencies.md
```diff
@@ -0,0 +1,112 @@
...
+8. Shift context of building docker images to root folder (otherwise we cannot share `requirements-common.txt` and `requirements-dev.txt` symlinks across subdirectories)
+9. Extend every `setup.py` to contain following entries (`extras_require` might contain more keys)"
+```python
+install_require=requirements + requirements_common
```
dorianj, line 62: For Databuilder, I think this is less-than-ideal, since we expect it to actually be installed as a library, and thus having precisely pinned deps in setup.py might cause issues. For the other "libraries", I don't think it's as a big a deal, since they are generally used standalone. wdyt?

rfcs/038-unify-shared-dependencies.md
mgorsk1: as far as I know (and from research I did) there is no way to make setup.py constraints aware. what you would do in such case is just list shared dependencies without pinning versions and then, when doing pip install you supply pinned versions using `--constraint`  parameter. what's important is that --constraint doesn't force installation of additional deps - if something is listed in constraints file but not in install_requires then it won't be installed

rfcs/038-unify-shared-dependencies.md
mgorsk1: I think git symlinks are supported image

rfcs/038-unify-shared-dependencies.md
mgorsk1: from my brief experiments I've actually identified databuilder as more `problematic` - mostly cause of the multitude of optional extractors (and hence > a lot of sections in `extras_require`). We could be good by partially removing it from the approach - so keep the requirements.txt as is, do not include `requirements-common.txt` but use shared `requirements-dev.txt`

rfcs/038-unify-shared-dependencies.md
mgorsk1: to be completely precise - I also was thinking about making clearing `setup.py` files out of deps and pull them by reading `requirements*txt` files

rfcs/038-unify-shared-dependencies.md
dorianj: nice

rfcs/038-unify-shared-dependencies.md
dorianj: sgtm to exclude it.

rfcs/038-unify-shared-dependencies.md
```diff
@@ -0,0 +1,125 @@
...
+Following original RFC:
+
+```
+Dependency Management: Python packages are pretty much the same for each proxy, which makes it hard to sync across all the above repositories. As a result most of the time, each proxy is running its own version of the dependency.
+A few examples:
+
+| Package             | amundsenmetadata   | amundsenfrontend  |
+| --------------      |:------------------:| -----------------:|
+| amundsen-common     | >=0.8.1            | ==0.6.0           |
+| flake8              | ==3.5.0            | ==3.8.4           |
```
Fokko, line 25: For test libraries as flake8 I would highly recommend using something like [pre-commit](https://pre-commit.com). This makes it very easy to do the checks, and isolate the test libraries like flake8, mypy, black, etc from the rest. This would also remove the issue of test-dependencies clashing because pre-commit will create a separate environment per hook. Updating would be quite straightforward, using `pre-commit autoupdate`.

Fokko (commented):
I think this is a great idea. I have one suggestion by splitting out the test-dependencies using pre-commit. LMYWYT

rfcs/038-unify-shared-dependencies.md
mgorsk1: I agree that pre-commit could be a good fit. Would you then replace `make test` in ci/cd with `pre-commit run --all-files` ?

Fokko: I think triggering `make test` would be a nice transition. Check if `pre-commit` is on the PATH, and if not, give the developer instructions on how to install this. I agree that this might be bigger than the RFC. I don't want to blow up this RFC :)

mgorsk1: I'll extend the list of future possibilities with this suggestion then!

rfcs/038-unify-shared-dependencies.md
```diff
@@ -0,0 +1,126 @@
...
+1. Split `requirements.txt` for each subdirectory into `requirements.txt` and `requirements-dev.txt` (there is no need to install flake8 when doing for example `pip install amundsen-metadata` which is currently the case)
+2. Identify libraries that are shared across all (or majority - see `Drawbacks`) of subprojects
+3. Create `requiremnts-common.txt` file in root folder of amundsen repo containing pinned dependencies for shared libraries
+4. Remove libraries listed in `requirements-common.txt` from `requirements.txt` of each subdirectory
+5. Create `requirements-dev.txt` file in root folder of amundsen repo containing pinned dependencies for shared dev libraries (mypy, flake8, etc.)
+6. Remove libraries listed in `requirements-dev.txt` from `requirements.txt` of each subdirectory
+7. Create symbolic link to `requirements-common.txt` and `requirements-dev.txt` in each subdirectory
+8. Shift context of building docker images to root folder (otherwise we cannot share `requirements-common.txt` and `requirements-dev.txt` symlinks across subdirectories)
```
sewardgw, line 65: I think the idea of shifting the build context up one directory makes sense, it also gives you the ability to copy in the `amundsen common` that is on the same brach (not in pypi) and copy it into each image. This could also give the benefit that we wouldn't need to add new versions of `amundsen-common` to pypi.

sewardgw (approved):

rfcs/038-unify-shared-dependencies.md
mgorsk1: How would we go about pypi installation of amundsen services if common is not a pypi pckage? I think not all users use pre-built images. 

Moreover, we have introduced apache atlas utils recently as a part of common and will be relying on them for both databuilder and metadata.

verdan:
@mgorsk1 how would that work with the PyPi versions of Amundsen packages?
PyPi requires the files needed for the setup.py in the build, that means we need to include the `requirements-common.txt` and other common requirements file within the build for each package.

rfcs/038-unify-shared-dependencies.md
sewardgw: Is the pypi installation of search, metadata and frontend really that common of use-case? I get that we wouldn't want to pull the plug on that user group, and by no means am I against keeping it in pypi and just overriding it during the build process, but it just feels like one of those areas where we _shouldn't need_ to have the common directory in pypi any more since it's now in the same branch - mostly thinking out loud here.

mgorsk1:
That's exactly how it would work. The requirements-dev and requirements-common files would be a part of artifacts built by sdist and bdist. It should be possible with proper manifest definitions (assuming your concern is over symlinks resolution?)

rfcs/038-unify-shared-dependencies.md
```diff
@@ -0,0 +1,126 @@
+- Feature Name: Unify shared dependencies
+- Start Date: 2021-05-22
+- RFC PR: [amundsen-io/rfcs#38](https://github.com/amundsen-io/rfcs/pull/38) (after opening the RFC PR, update this with a link to it and update the file name)
```
Golodhros, line 3:
```suggestion
- RFC PR: [amundsen-io/rfcs#38](https://github.com/amundsen-io/rfcs/pull/38)
```

feng-tao:
how about databuilder? currently the model is defined in both common and databuilder which we should eventually make databuilder depends on common model to make it consistent to avoid model drifting.

mgorsk1:
In principle I agree that long term there should not even be models folder in any of the project except for the common. This would be however a major refactor from a different angle than I had in mind for this RFC. I will add this to future possibilities.

verdan (approved):
this looks good to me and really excited about this 🙂 
I don't know the technical challenges regarding the PyPi release yet and adding links to files outside the root directory, but more than happy to at least give this a try. 
The next step would be to remove all the local models from each package and simply rely completely on common models.

mgorsk1:
Nice, thanks! I've already started to draft something under https://github.com/amundsen-io/amundsen/pull/1163 , seems most major things are resolved - mostly flake8 complains since I've bumped the version and not all the code is now compliant.

PR #39, opened 2021-05-25
Remove double relationships RFC

gorbachev:
Signed-off-by: Tero Paananen <tpaananen@bloomberg.net>


rfcs/000-remove-double-relationships.md
```diff
@@ -0,0 +1,102 @@
...
+
+Currently all (most?) relationships between Neo4j nodes are modeled with
+two directional edges. This RFC proposes we only use one directional edge
+between the nodes.
+
+## Motivation
+
+The double relationship graph model makes Neo4j queries more complicated
+than they need to be, with no performance or functional benefit. Additionally
+due to conflicting changes in the past (PR 70 and PR 206), Amundsen instances
```
dorianj, line 18: Can you link to these PRs?

rfcs/000-remove-double-relationships.md
```diff
@@ -0,0 +1,102 @@
...
+- a thread where I asked the question on why the double relationships exist
+https://app.slack.com/client/TGFR0CZM3/CGFBVT23V/thread/CGFBVT23V-1563487877.254000
+- a thread with Joshua Haskins' benchmarking results for modeling the relationships
+  with one or two edges
+
+## Transition Path
+
+The implementation work involved:
+
+For each double relationship pair, we need to decide which relationship type
```
sewardgw, line 35: Are you thinking there is a way to apply this generically? Perhaps asked more broadly - what else may need to be done by companies that have created their own models?

rfcs/000-remove-double-relationships.md
gorbachev: Done

rfcs/000-remove-double-relationships.md
gorbachev: Good point. I wasn't thinking of custom models at all. I'm not sure what the right approach would be for those.

Maybe offer guidance on how to make corresponding changes on their own code, if needed? Then leave it at their own discretion whether they want to do the change.

rfcs/000-remove-double-relationships.md
```diff
@@ -0,0 +1,103 @@
...
+There is no transition path for new installations. They will simply start using
+the single relationships graph model from the get go.
+
+For existing Amundsen instances we'll need to make sure the database migration
+script is run after upgrading Amundsen to a version where the double relationships
+have been removed.
+
+## How We Communicate This
+
+TBD
```
sewardgw, line 60: While communication to the community should help to mitigate any issues, it seems like one of the largest potential risks (IMHO) could be having a version mismatch b/w the data (largely impacted by data builder) and the metadata service. In addition, there isn't a hard requirement that links databuilder and metadata today that we could rely on to enforce version compliance.

Do you have thoughts on the feasibility on making the metadata layer backwards compatible and how the rollout of this could be "accident proof" across services to de-risk any implementation. ?

rfcs/000-remove-double-relationships.md
alexbrisan: My 2CS here is that if you’re deviating from the standard datamodels (although I *think* it’s only an issue if you modify, rather than add) then that’s kind of you. What we can do to help is create some very clear documentation of what the data models are, rather than having to extract them from source code.

rfcs/000-remove-double-relationships.md
alexbrisan: I think if we freeze the metadata layer as it is now, and factor out the model classes that are defined separately in databuilder and metada/search into a common library, we can rely on that ancestor as the source of truth, and use that to keep things in sync. A good place to start is amundsenrdns IMHO

sewardgw: I'm largely aligned here and, should this approach be taken in the RFC, would like to see the next level of detail on the how to do this in the RFC.

rfcs/000-remove-double-relationships.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Summary
+
+Currently all (most?) relationships between Neo4j nodes are modeled with
+two directional edges. This RFC proposes we only use one directional edge
+between the nodes.
+
+## Motivation
+
+The double relationship graph model makes Neo4j queries more complicated
```
feng-tao, line 16: could you describe this complicate part in more detail? isn't similar to normal rdbms FK and backreference between two models?

Also what is the benefit besides no need to specify two relationships instead of one?

@jinhyukchang done some initial perf testing for duel relationship initially FYI

dorianj: > Also what is the benefit besides no need to specify two relationships instead of one?

I think if there's no performance degradation, and the query expressibility isn't impacted (i.e. it shouldn't be challenging to write the same query as before), then it's purely cruft, no?

feng-tao: I am not sure if it is cruft as benchmark is always subject to lots of dimensions. I think the key of this rfc requires to layout the detail of neo4j data migration if we want to go from duel relationship to one.

rfcs/000-remove-double-relationships.md
gorbachev: I'm submitting some changes to the RFC in a minute to discuss how this already happened in PR 70 referred in the RFC.

Personally I'm of the opinion the problem you're asking about should be addressed separately. It might be beneficial to address it before working on implementing this RFC, though.

rfcs/000-remove-double-relationships.md
gorbachev: I think cruft is a good way of describing it. It's not really breaking things, it's just sub-optimal, and a bit of a bad design.

There are performance implications, however. Not that you would really notice unless the size of your graph was very, very large. I'm not sure how large the biggest Amundsen deployments are in terms of number of nodes and edges in the Neo4j database, but I'd be surprised if it hit anywhere near where Neo4j performance would actually start suffering due to this issue. Obviously you could also just throw more hardware on it anyway, as implied by @feng-tao.

I'm submitting some changes to the RFC to outline this as soon as I can write it out. Hopefully in an hour or two.

gorbachev: Additionally, no, this is not the same as back references in an ORM / relational database. Back references are the most efficient way an ORM framework can effectively refer back to a parent object.

There is no need to create the bi-directional relationships in Neo4j, because you can follow the one relationship without any issues on performance, functionality or impact on maintainability.

The following two Cypher queries return the exact same results.

`
MATCH (n1:TYPE_A)-[r:REL]->(n2:TYPE_B) RETURN n1, r, n2
MATCH (n2:TYPE_B)<-[r:REL]-(n1:TYPE_A) RETURN n1, r, n2
`

PR #40, opened 2021-06-08
chore: Updates RFC process to tweak RFC proposals with UI changes

Golodhros:
# Description
Adds comments about UX/UI changes on RFCs
Updates some RFC file names


README.md
```diff
@@ -59,16 +59,18 @@ In short, to get a major feature added to Amundsen, one must first get the RFC m
...
+   - This step is taken **when enough of the tradeoffs have been discussed** so that the maintainers are in a position to make a decision. That does not require consensus amongst all participants in the RFC thread. However, the argument supporting the disposition on the RFC needs to have already been clearly articulated. There should not be a strong consensus against that position outside of the maintainer's group.
+   - The FCP period lasts **seven days**, and the beginning of this period will be signaled with a comment and tag on the RFC's pull request.
+   - For RFCs with lengthy discussion, the motion to FCP is usually preceded by a summary comment trying to lay out the current state of the discussion and major tradeoffs/points of disagreement.
+   - Maintainers will ping the `#amundsen` channel about the RFC to attract the community's attention.
+   - In case the RFC contains a UX/UI change, the RFC champion will reach out to any available Design resource to see if we can get some help on the visuals.
```
dorianj, line 66: I'd like to scope this smaller -- to basically just list it in the default template. I think it's reasonable to assume that designers can follow the #rfcs channel just the same as interested engineers do. WDYT?

templates/000-feature-template.md
```diff
@@ -27,7 +27,7 @@
 
 > Explain the UI changes that your proposal would need (if applicable). This could mean:
 
-> Provide a high level diagram or wireframe about the change.
+> Provide at least a low-fidelity wireframe of the change.
```
dorianj, line 30: 👍

templates/000-feature-template.md
```diff
@@ -27,7 +27,7 @@
...
 > If applicable, suggest error, empty and loading states for the change.
```
dorianj, line 31: I don't think we're actually doing any RFCs with this level of specificity -- maybe to simplify, we simply remove this so that the bulleted list is gone and the requirement is clear?

README.md
Golodhros: Well... I don't think Knowl is checking that channel. I think it is easier for the RFC champion to reach whenever is necessary, as it shouldn't be too often.

templates/000-feature-template.md
Golodhros: Good call, removing it.

README.md
dorianj: I think Knowl should check out that channel, it should be low-volume enough to keep up with. I'm concerned that it'll be hard to make sure committers consistently do that (e.g. how are committers to know what design resources are available?). There are already parallels with specific piece of code, for example, that engineers make sure to keep an eye out for.

Golodhros: Sadly, Knowl is a bit overworked these days at Lyft.

Oh, I wouldn't expect the committer to do that, only the maintainer, and only when we get into the FCP.

dorianj: but not all of the maintainers will know about what design resources are available, either (since they're not all at lyft, and it's definite in the future there will be more designers than just Knowl). Perhaps a Lyft PM could keep an eye on RFCs and escalate them?

I'm concerned that if Knowl herself isn't able to keep an eye on this (which is totally understandable and not a problem), that she wouldn't have the bandwidth to create the designs, either (especially in a timely manner -- I don't think we should block any new functionality landing in the project on design).

kbaekgit: Thanks for this addition- and I'll try my best to check the #rfc channel ~weekly, which seems to be the frequency of posts and since there is a 1wk period for feedback. Would appreciate you guys alerting me in case I miss something, and please encourage committers to include any available visuals for UI changes!

Golodhros: OK, removing this then!

PR #41, opened 2021-07-19
Table checks frontend component

danwom:
Adds an RFC to propose:

-a frontend component for displaying 3-valued table status: number of checks, number passing, and an optional timestamp
-stubbing of the backend for this component


rfcs/039-table-data-quality-checks.md
```diff
@@ -0,0 +1,61 @@
...
+## Drawbacks
+This can be seen as a somewhat shallow integration that doesn't allow for some deeper integrations. Granted that we are
+trying to support a generic model, it's difficult to support a deeper integration given that we are not endorsing a 
+data quality service provider.
+
+## Alternatives
+There are several alternatives: 
+  - Ingest data quality checks as part of the databuilder and display the status as part of the table metadata API. 
+The displayed status in this case may be stale.
+  - We could build data quality checks as an Amundsen service and provide a much deeper level of service, but this
```
dorianj, line 60: I'm similarly negative on this: integration is the way here

feng-tao: +1, we should link dq service that is hosted externally instead of putting DQ inside amundsen which is more for discovery purpose.

rfcs/039-table-data-quality-checks.md
```diff
@@ -0,0 +1,61 @@
+- Feature Name: table_data_quality_checks
+- Start Date: 2021-07-16
+- RFC PR: [amundsen-io/rfcs#XX](https://github.com/amundsen-io/rfcs/pull/XX)
+# Table Data Quality Checks
+
+## Summary
+This RFC will define a generic interface for accessing data quality checks in Amundsen. If available, we aim to expose
```
sewardgw, line 7: I think it would be beneficial to take a slightly more comprehensive approach where we define the underlying data models that Amundsen can store internally for data quality checks, even if those are not immediately implemented. The thought here being that if we take the approach where we only define an interface for accessing data quality checks (presumably from an external system) it could possibly constrain the approach (or cause significant refactor) that we eventually take for storing similar data within Amundsen. 

I'd like to see us look at this from a slightly broader perspective to ensure that the design will be highly reusable, especially considering that there are several major players in the DQ space.

rfcs/039-table-data-quality-checks.md
sewardgw: +1 for OPs statement.

In regards to:

> we should link dq service that is hosted externally instead of putting DQ inside amundsen which is more for discovery purpose.

There are some dq services that are batch processes which produce a static output in JSON, html, etc.  Are you suggesting that Amundsen should not coordinate and execute the dq tests _or_ is this comment saying that Amundsen should not ingest the output from the tests that provide table/column level quality results?

rfcs/039-table-data-quality-checks.md
danwom: I see your point about the value of a deeper integration with reusable and generic models.

One caveat about DQ checks is that live data is highly valuable compared to stale data that was ingested in a databuilder job some hours ago. For this reason we think accessing an external API that can provide live data would be preferable for this integration. 

We could create our own model for DQ checks in Amundsen and then pull updates from a service periodically, or on demand, but I'm not sure how much value that would bring for the current use case.

rfcs/039-table-data-quality-checks.md
danwom: I think we're saying Amundsen should not be the entity responsible for registering and running the DQ tests. We should be integrating with an external DQ check service that can provide this information.

feng-tao:
lgtm

rfcs/039-table-data-quality-checks.md
```diff
@@ -0,0 +1,61 @@
...
+the status of these data quality checks in the Amundsen Table Details page.
+
+
+## Terms
+- Data Quality Check - Like unit tests for code, data quality checks are assertions that try to guarantee certain 
+  characteristics of data. For example a data quality check could be as simple as asserting that a column is not null,
+  or implement some complex custom logic.
```
mgorsk1, line 14: Do you already have in mind some tools from which these quality checks could have been collected? I can think of Great Expectations and dbt tests but would be useful to have broader list. It's also something @sewardgw had good point about the interface to be generic enough we are not forced to do backwards incompatible changes later on.

From what I can see the mockups you introduced somewhat adhere to GE format but I am not sure about dbt (cause I don't have exp with it).

rfcs/039-table-data-quality-checks.md
```diff
@@ -0,0 +1,61 @@
...
+
+## UX Explanation
+A new section will be added to the left side of the Table Details Page.
+
+![Data Quality Preview](../assets/039/data_quality.png)
+
+## Technical Details
+Table Data Quality Checks will be implemented as an optional feature on Amundsen Frontend. This feature will be hidden 
+behind a config flag and will use a generic client much like the existing data preview client. The generic client should
+implement the following interface:
```
mgorsk1, line 36: Any reason why we are not considering extending Table model with quality check entity?

I think there are a lot of tools that can be used to run DQ checks but don't provide integration in the form of API to query those results and this approach limits us to only tools that would have it.

rfcs/039-table-data-quality-checks.md
mgorsk1: I think another approach here could be to provide an endpoint that those quality services could call and send latest DQ check info. As stated in another comment, I think there are a lot of tools that can be used to run DQ checks but don't provide integration in the form of API to query those results and this approach limits us to only tools that would have it.

I also disagree with the point that databuilder is not a good fit here. A lof of DQ unit tests might be run with scheduling service like Airflow and it could be downstream task to ingest the data to Amundsen after running DQ test. The delay of 1-2 minutes is not something I would consider to be an issue.

In our case we run unit tests with Great Expectations using Airflow and if Amundsen enables DQ checks surfacing I would just register a callback ingesting this live to Atlas using databuidler job.

rfcs/039-table-data-quality-checks.md
danwom: I am not against extending the Table model and defining databuilder tasks to ingest DQ checks from various providers. I think that's a great idea with many merits.

However this RFC isn't intended to define a fully-fledged solution, but rather proposes a simple integration via an API interface. 

That being said I have [this old PR](https://github.com/amundsen-io/amundsen/pull/1073) which was intended to be merged into a private repo, until I was asked to contribute it to OSS. So perhaps this solution is a little narrow minded to retrofit an existing PR.

rfcs/039-table-data-quality-checks.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Motivation
+While Amundsen is not currently building unit tests for data, we can empower our users by allowing integrations 
+with a generic external data quality service. Exposing the status of data quality checks in Amundsen can give users 
+an at-a-glance summary of the table's quality. This can help a user to quickly decide if they want to use this table.
+
+## Product Details
+Users will see a new section on the left panel of the Table Details page titled `Data Quality Checks`. There will be
+a message and icon for the current status of these checks, if they exist. A passing table might say 
+`✅ 10 of 10 checks passed` and a failing table might say `❌ 3 of 10 checks failed`. Additionally, the user may see
```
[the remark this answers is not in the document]
danwom: The way I interpret it is: Checks run on various cadences. Some tests run hourly, some might run daily or even weekly. For each individual check, how many of the last run passed or failed.

This interpretation is left up to the specific implementation. We define an API with fields like `num_checks_passed`, `num_checks_total`, and `last_run_timestamp`. These can mean slightly different things if you wish.

rfcs/039-table-data-quality-checks.md
sewardgw: Too add to this, Amundsen is well positioned to be able to propagate data quality downstream using the lineage information available. While I think that it is fine to enrich the data displayed in the UI using an external API to retrieve some more in-depth information, Amundsen should be able to take the information that it knows about failed DQ tests and do something with that information on downstream assets (e.g. place an alert on affected dashboards, etc.).

dkunitsk:
@mgorsk1 @sewardgw Thanks for your thoughts above. Your question around comparing the API vs offline databuilder approaches makes a lot of sense. 

At this time, we are not ready to add full first class data quality checks to Amundsen. We aren't even ready to put forward a good generic definition of a check, as it might have a lot of dimensions and we're not ready to pin it down (condition, time range, columns validated, correctness/failure percentage, severity, owner, etc). This would be a great direction, but not what we were planning to tackle.

Instead, we're proposing a simple "Table Checks" [frontend component](https://github.com/amundsen-io/rfcs/blob/dwon-data-quality-checks/assets/039/data_quality.png) to display limited pass/fail state. Originally, we were going to power this component with a customizable client (as this follows some prior art in Amundsen) but given your feedback we have an alternate suggestion:

> Add a placeholder endpoint (`/api/quality/v0`) in the frontend service which returns a dummy response (e.g. "0 out of 0 checks successful"). Companies who have an API to hook up to this frontend component could override the endpoint implementation with a call to their API. Otherwise, this state would be a clean starting point for adding backend ingestion down the line. 

This has the advantage of moving in the general correct direction (a pluggable "checks" frontend component) without making premature decisions since we'd be starting with the the most bare-bones definition possible.

rfcs/039-table-data-quality-checks.md
feng-tao: DQ normally requires low sla which requires a push model while Amundsen / databuilder still has a pull model approach. How could the downstream consumer depends on Amundsen metadata in this case?

sewardgw: I think that databuilder can really be considered push or pull based on the context in which it is executed. For example, for data quality, a company may have several teams that each use airflow to orchestrate the DAGs that do something like:

- Task 1: load a table
- Task 2: run DQ checks
- Task 3: push DQ results to Amundsen

Push / pull semantics aside, Amundsen would receive the data as soon as it is available. The specific cadence for which this DQ process runs actually does not matter, whether it is hourly, daily, weekly, etc. As soon as Amundsen receives an update that the DQ checks have failed it should take action to propagate that information downstream.

In regards to the SLA, from both the user perspective, as well as Amundsen's perspective, the state of the data quality is the last known value that was generated, regardless of whether or not the actual underlying data has changed. Therefore, the actual SLA for DQ is not important, what is important is that DQ events are made available to Amundsen as soon as they are created.

mgorsk1:
What I am missing about this proposal is some example/guidance on what DQ systems you have in mind to be used with current approach @dkunitsk Is it something custom you have in Lyft? Having such list enables us to expand `Future possibilities` section  with more guided way to build on top of this. 

I agree with @sewardgw on SLA and push/pull semantics and that either way it's possible to have data in Amundsen as soon as it's possible.

dkunitsk:
@mgorsk1 unfortunately we don't have most of the answers at this time. You're very right that a big portion of data quality will be at the column level, and we will be working in that direction in the near future. 

At the moment we're proposing only:
- a frontend component for displaying 3-valued table status: number of checks, number passing, and an optional timestamp
- stubbing of the backend for this component

In some sense, this is un-opinionated enough that one could crudely plug in any DQ system. And we believe it's a generic starting point for all generalizations discussed - databuilders, push/pull questions, check definition, etc.

Given the good feedback above, we have a much better grasp of what the community is looking for in a generic DQ check feature and we'll be tailoring future proposals according to this feedback. 

In the meantime, we'll adjust this RFC to make our scope more clear so we don't get stuck in the deeper questions just yet.

rfcs/039-table-data-quality-checks.md
dkunitsk: These are wonderful questions, but as I've tried to point out below they are far outside the scope of what we were proposing - which is on us for not being as precise as we should have been. Please take a look at my thoughts here: https://github.com/amundsen-io/rfcs/pull/41#issuecomment-892064322

We'll iterate on the RFC to make its intention more clear. 

In the meantime, our proposed work will take no direction one way or the other on the above.

rfcs/039-table-data-quality-checks.md
dkunitsk: We are not building this with a particular tool in mind. At the 1M-mile view, we think our proposed definition is so generic that it could crudely fit anything. The idea is to start here and expand once people start plugging in particular systems. 

At Lyft we do have a DQ system we're planning to plug in, but it exposes a much more complicated API and semantics which we are not taking into account here. Instead, we asked ourselves what would be the most generic first step.

rfcs/039-table-checks-frontend-component.md
```diff
@@ -0,0 +1,60 @@
...
+}
+```
+
+## Drawbacks
+Granted that we are trying to support a generic model, it's difficult to support a deeper integration given that we 
+are not endorsing a data quality service provider.
+
+## Alternatives
+There are several alternatives: 
+  - Ingest data quality checks as part of the databuilder and display the status as part of the table metadata API.
```
dkunitsk, line 57: Note: this is not so much an alternative but rather an addition.

dkunitsk (approved):
+1. After the pared down scope, this leaves room for multiple directions on the unsolved questions (push/pull, specific integrations, etc).

feng-tao (approved):

PR #42, opened 2021-07-28
Editable programmatic descriptions

jkulzick:
Adds an RFC to propose editable programmatic descriptions


feng-tao:
cc @mgorsk1 @sewardgw @allisonsuarez @Golodhros could you all take a look first? I am packed this week.

rfcs/042-editable-programmatic-descriptions.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Reference-level Explanation (aka Technical Details)
+
+The implementation will require:
+- Updates to the frontend and metadata service to support editing programmatic descriptions via the API
+- Updates to the frontend to display editable programmatic descriptions in the table details page 
+
+### Metadata
+
+- Add an optional "description_source" parameter to the existing put_table_description API endpoint that defaults to None.
```
sewardgw, line 47: Is `description_source` being used to determine whether or not a prog desc can be edited? If so, what are some potential values here and is there a need to have a well-defined, limited set of consistent values?

rfcs/042-editable-programmatic-descriptions.md
```diff
@@ -0,0 +1,100 @@
...
+
+This feature allows users to configure their Amundsen installation to support one or more editable programmatic descriptions. These editable programmatic descriptions allow users to enrich the metadata displayed on the table details page with additional business specific human generated information. This feature will allow them to add/edit these programmatic descriptions directly in the UI.
+
+## Guide-level Explanation (aka Product Details)
+
+An editable programmatic description is the same as a regular programmatic description, except that it is maintained directly in the UI by a human instead of by a scheduled databuilder job. For example, users could configure an editable programmatic description with a source of data_usage_policy and then update the data_usage_policy on individual table detail pages with a link to the applicable data usage policy on their company's content management site.
+
+Similar to table descriptions, if a table is disabled for editing, then so are the editable programmatic descriptions.
+
+Databuilder jobs may still be useful on a one-off (i.e. non-scheduled basis) to seed Amundsen with manually curated data or perform one-off data conversations.
```
sewardgw, line 24: Are you stating that databuilder will be the _only_ way to create these editable prog descs (where as once they are created they can be edited in the UI)? 

Also, can you quickly highlight the change to databuilder as well, in particular the prog desc model?

rfcs/042-editable-programmatic-descriptions.md
jkulzick: The `description_source` would be used to populate the `source` attribute of [ProgrammaticDescription](https://github.com/amundsen-io/amundsen/blob/45f77100cd13c4faa4f922e1daf5e2e92fd23434/common/amundsen_common/models/table.py#L118). Similar to how databuilder currently works for loading descriptions, if the `description_source` is None (the default value) then the `description` populate the table's Description, otherwise the `description` and `description_source` populate a Programmatic_Description. I opted for `description_source` since it looks like we already have a parameter called `source` that is used to denote where the source of the update (see [here](https://github.com/amundsen-io/amundsen/blob/45f77100cd13c4faa4f922e1daf5e2e92fd23434/frontend/amundsen_application/api/metadata/v0.py#L311) and [here](https://github.com/amundsen-io/amundsen/blob/45f77100cd13c4faa4f922e1daf5e2e92fd23434/frontend/amundsen_application/static/js/ducks/tableMetadata/api/v0.ts#L103)).

rfcs/042-editable-programmatic-descriptions.md
jkulzick: TL;DR - users can add programmatic descriptions to a table via the UI (if in `PROGRAMMATIC_DISPLAY` as editable) and via datdabuilder (if they decide on their own to do that). the current proposal requires no changes to databuilder.

Once a programmatic description is configured as editable in the configuration of the frontend service in `PROGRAMMATIC_DISPLAY` then it will always appear in the UI on every table detail page with placeholder text. So once that happens, users can edit the editable programmatic description to add a value if one does not exist.

AFAIK databuilder currently allows users to add any programmatic description they desire, even if it's source is not configured in `PROGRAMMATIC_DISPLAY`. Of course, it will only be displayed in the UI if it is in `PROGRAMMATIC_DISPLAY`. Based on this understanding, there should be no changes required to support a one-off load via databuilder of editable programmatic descriptions.

rfcs/042-editable-programmatic-descriptions.md
```diff
@@ -0,0 +1,100 @@
...
+
+There will be no impact to existing programmatic descriptions. They will remain non-editable via the UI unless users intentionally update the existing configuration in their Amundsen installation to make them editable.
```
mgorsk1, line 26: Does it mean editable programmatic descriptions won't be distinguishable from those loaded by databuilder?

rfcs/042-editable-programmatic-descriptions.md
```diff
@@ -0,0 +1,100 @@
...
+
+## UI/UX-level Explanation
+
+Editable programmatic descriptions would be displayed alongside programmatic descriptions in the same way they are displayed today. They will be rendered as markdown, located in one of three sections (left, right, other) in the page and sorted according to display_order.
+
+There are two differences between editable and non-editable programmatic descriptions:
+1. They are editable (if the table is editable)
+2. They will always be displayed. If there is not a value then placeholder text will be displayed instead.
+
+_**Note:** The pink text is only visible in the mocks to call out design details and changes._
```
mgorsk1, line 36: this is a question for the UI proposal - how will existing Amundsen users (not the team that is deploying Amundsen) `know` that given programmatic description is editable if `edit` icon just appears on hover?

rfcs/042-editable-programmatic-descriptions.md
```diff
@@ -0,0 +1,100 @@
...
+
+Similar to table descriptions, if a table is disabled for editing, then so are the editable programmatic descriptions.
+
+Databuilder jobs may still be useful on a one-off (i.e. non-scheduled basis) to seed Amundsen with manually curated data or perform one-off data conversations.
+
+There will be no impact to existing programmatic descriptions. They will remain non-editable via the UI unless users intentionally update the existing configuration in their Amundsen installation to make them editable.
+
+## UI/UX-level Explanation
+
+Editable programmatic descriptions would be displayed alongside programmatic descriptions in the same way they are displayed today. They will be rendered as markdown, located in one of three sections (left, right, other) in the page and sorted according to display_order.
```
mgorsk1, line 30: afaik currently the order is alphabetical, so this RFC also proposes custom prog description sorting?

rfcs/042-editable-programmatic-descriptions.md
jkulzick: Is the question about the scenario where an implementor chooses to seed an editable programmatic description via databuilder and then edit the value of the editable programmatic description for a given table(s) via the UI? If so, then yes, you cannot tell if the value came from databuilder or the UI.

Please let me know if you have specific concerns that you think we should address in this RFC.

We could further lock this down by requiring the `PROGRAMMATIC_DISPLAY` configuration in the databuilder service and the metadata service so that databuilder can only load programmatic sources defined in the configuration and metadata service can only update editable programmatic sources defined in the configuration.

Currently, databuilder appears to allow users to load arbitrary sources as programmatic descriptions that may have nothing to do with what is defined in `PROGRAMMATIC_DISPLAY`.

rfcs/042-editable-programmatic-descriptions.md
jkulzick: That is a good point. We should always display the edit icon. This is the default behavior of `EditableSection` today. I will update the mock.

rfcs/042-editable-programmatic-descriptions.md
jkulzick: My proposal would be to keep the sort order the same as it is today. My assumption was that this [function](https://github.com/amundsen-io/amundsen/blob/c4c908df7a14e72058932f4d597656df5a926f48/frontend/amundsen_application/api/utils/metadata_utils.py#L235) sorted the programmatic descriptions of a table according to the `display_order` configured in `PROGRAMMATIC_DISPLAY` (if it was configured). But it sounds like that sorting is not maintained from the API layer to the frontend.

I will update the RFC to call out that sorting will not change.

rfcs/042-editable-programmatic-descriptions.md
jkulzick: Updated

rfcs/042-editable-programmatic-descriptions.md
jkulzick: Updated

feng-tao:
there are active discussion, any reason we move it to FCP?

rfcs/042-editable-programmatic-descriptions.md
```diff
@@ -0,0 +1,100 @@
...
+- RFC PR: [amundsen-io/rfcs#42](https://github.com/amundsen-io/rfcs/pull/42)
+- Amundsen Issue: [amundsen-io/amundsen#0000](https://github.com/amundsen-io/amundsen/issues/0000)
+
+# Editable Programmatic Descriptions
+
+## Summary
+
+Programmatic descriptions are a useful way for Amundsen users to store and display business specific metadata about their tables in Amundsen. This feature allows for some customization of Amundsen without customizing the underlying software. Currently, this feature focuses on the machine generated metadata use case. This type of data can be extracted from the source system of truth and ingested into Amundsen via databuilder.
+
+The same need exists for business specific human generated metadata. This type of data does not exist in a machine generated form that can be easily extracted from a source system. Instead, users need to add/edit this metadata directly in the Amundsen UI.
```
feng-tao, line 12: for the case that programmatic description are machine generated vs manual edit, how do you enable/disable the edit for each?

feng-tao: ok, saw it in the later section that you config through FE config.

rfcs/042-editable-programmatic-descriptions.md
feng-tao: I have the same qq with @mgorsk1 as how you identify it is manual edit vs daily refreshed from databuilder.

And for manual edit programmatic description, how do you avoid the value override by databuilder ETL?

feng-tao: Just FYI I am not sure sharing the config between FE and databuilder is a good idea.

jkulzick: @mgorsk1 @feng-tao I definitely understand the concern, but I'm curious what your recommendation for resolving this would be given the product as it stands today? How is this concern addressed for other features that are editable via both databuilder and the UI today (e.g., table descriptions, table owners)?

If there is a standard that is adhered to in the product then I would also like to apply it here as well. Please let me know.

feng-tao: For other cases, you could take a look at https://www.amundsen.io/amundsen/faq/#how-to-avoid-certain-metadatas-in-amundsen-got-erased-by-databuilder-ingestion which the config is on node label level. But for this case, I think we need to think about how to handle it in a hybrid way.

feng-tao: But this concern needs to be sort out otherwise the proposal won't work as expected.

jkulzick: @feng-tao Here are the options I came up with. Let me know what your thoughts.

I did not include an option of having some config in databuilder that lists the editable programmatic descriptions since you mentioned that sharing the config is not a good idea.

1. Use new node label for human-generated descriptions

E.g., `Manual_Description` or `Editable_Description`

Under this proposal, human-generated descriptions would be stored separately from programmatic descriptions. This would allow for the use of the [label level solution](https://www.amundsen.io/amundsen/faq/#how-to-avoid-certain-metadatas-in-amundsen-got-erased-by-databuilder-ingestion) for preventing unintended updates via databuilder.

Of course, this would require more implementation work (e.g., databuilder model, update to table metadata common model, etc.). Also, separating storage for these two concepts just because one is editable does not seem ideal.

2. Prevent databuilder from updating nodes last updated by the UI

Introduce an optional node property `last_updated_by_ui` that for the time being could be used only for programmatic descriptions.

When the UI updates editable programmatic descriptions, it would set the property `last_updated_by_ui` on the `Programmatic_Description` node to `true`. We could update the merge statement in [create_node_merge_statement](https://github.com/amundsen-io/amundsen/blob/50cf839782f233ef30c4e4a3c0a280fc50583eab/databuilder/databuilder/publisher/neo4j_csv_publisher.py#L271) to reference this.

```
{% if update %}
MATCH (node:{{ LABEL }} {key: $KEY})
WHERE COALESCE (node.last_updated_by_ui, false) <> TRUE 
SET {{ PROP_BODY }}
{% endif %}
MERGE (node:{{ LABEL }} {key: $KEY})
ON CREATE SET {{ PROP_BODY }}
```

This would prevent databuilder jobs from updating editable programmatic descriptions. The only limitation is that there is no way to opt into updates via databuilder if you want them.

3. Option 2 + allow users to force update nodes last updated by the UI in databuilder

`NEO4J_FORCE_UPDATE_NODES` would be an array of node labels. Users would specify a node label in the array if they wanted databuilder to update a the node even if the UI was the last to update it.

This would be exposed in `Neo4jCsvPublisher` similar to the `self.create_only_nodes` and `self.is_create_only_node` pattern.

The revised merge statement would be:
```
{% if update %}
MATCH (node:{{ LABEL }} {key: $KEY})
{% if not force_update %}
WHERE COALESCE (node.last_updated_by_ui, false) <> TRUE 
{% endif %}
SET {{ PROP_BODY }}
{% endif %}
MERGE (node:{{ LABEL }} {key: $KEY})
ON CREATE SET {{ PROP_BODY }}
```

feng-tao: qq for @mgorsk1 , how does programmatic description work in atlas backend? And does the create-only node concept exist in other proxies?

rfcs/042-editable-programmatic-descriptions.md
Golodhros: I wonder if it would make sense to make these "programmatic descriptions", as they wouldn't be "programmatic" (except the seeded ones).

I think these could be called Human Generated Descriptions or a similar name.

Golodhros:
IMHO this feature makes sense. 

I would make it a different thing than the "Programmatic description", but I understand that it might be convenient to reuse the logic.

PR #43, opened 2021-09-22
Search revamp

dkunitsk:
Signed-off-by: Dmitriy Kunitskiy <dkunitskiy@lyft.com>

### DIFF
--- a/rfcs/000-search-revamp.md
+++ b/rfcs/000-search-revamp.md
@@ -1,0 +1,124 @@
+- Feature Name: Search Revamp
+- Start Date: 2021-09-20
+- RFC PR: todo
+
+# Search Revamp
+
+## Summary
+
+Revamp search with a cleaner backend implementation, more robust queries, better customizability, and
+making use of modern v7 ES features.  
+
+## Motivation
+
+At Lyft, search enhancement are the #1 most frequent Amundsen feedback. This is no surprise, as the current 
+technical state of search has not changed much from MVP.
+
+We see search as serving two distinct (but related) use cases:
+  - known-data searches, which have a starting point (e.g. tables with a `user_id` column in the `production` schema) 
+    and are more interested in listing the available data matching a known condition
+  - discovery searches, which are looking to find previously-unknown data that would be relevant for solving
+    one's problem
+
+Although we believe Amundsen has been particularly lacking in the discovery search category, our proposal is seeking
+to improve the experience of both search types. 
+
+
+## UI/UX Details
+
+Although the bulk of proposed changes will be to search result quality and relevance, we're also proposing
+a handful of UI/UX changes to filters. We're currently working on designs and will share those as soon as they're
+available.
+
+#### Multiple Values Per Filter.
+Allow multiple values (likely comma-separated) for filters, which has been very highly requested. 
+
+There is an obvious question of whether multiple values would be AND'ed or OR'ed together. Our goal is to provide
+additional expressiveness without going down the path of supporting full logical expressions. To better understand the problem,
+keep in mind that there's two kind of fields: single-valued fields (table name, schema name) and multi-valued fields (columns, tags, badges).
+Our proposed tradeoff between expressiveness and simplicity is to combine filter values for single-valued 
+fields with OR (e.g. "staging,production" filter applied to the "schema" field)
+whereas for multi-valued fields we'd give users an options to choose AND or OR 
+(a user would choose AND/OR to apply the filters "product_id,user_id" to the "column" field)
+
+#### Negation filtering
+Another highly requested feature is specifying negative filters - e.g. "schema is not temp, tmp, or temporary".
+As above, we want to avoid going down the rabbit hole of full-blown logical expressions. Likely, we'd implement
+this on the frontend as a separate set of filter boxes below the current set. Stay tuned for designs. 
+
+#### Type-ahead suggestions for filters
+Because values entered as filters are not interpreted or analyzed (this is the current behavior and is not expected to change),
+it can be hard for users to remember/guess the exact values to filter for. We'd like to provide typeahead
+suggestions that will suggest the most likely filter value given a prefix.
+
+
+## Backend Changes
+As part of this work, most of the elasticsearch [proxy](https://github.com/amundsen-io/amundsen/blob/main/search/search_service/proxy/elasticsearch.py)
+would be re-written, while keeping the API of the search service the same (in part to allow Atlas to stay unchanged).
+
+#### Upgrade Elasticsearch to v7
+Elasticsearch queries/mappings are generally written to a specific version, since a number of features are typically
+added and removed in each new release. We suggest moving to v7 as part of this revamp. How to gracefully handle 
+the backwards incompatibility for folks not ready to adopt a new ES version is an open question. 
+One option is to keep the existing ES proxy for v6 uses and add all our improvements in new code.
+
+#### Filtered vs unfiltered search unification
+Filtered search was added on top of regular search and was implemented separately via separate endpoints. 
+It also uses a different (and incorrect) elasticsearch [query_string](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html) command. 
+We plan to unify the implementations, so both use the same top-level elasticsearch command either with or without an additional filter specifier. 
+
+#### Replace JSON in queries/mappings with Elasticsearch library objects
+The requests made to Elasticsearch are constructed with either hardcoded (in the case of mappings) or templated JSON. 
+This makes it more difficult to compose and understand requests, 
+but more importantly it makes it very difficult to extend OSS code. 
+The Elasticsearch-dsl library includes classes for constructing requests by putting together native python objects.
+
+## Ranking/relevance changes
+The bulk of improvements we want to make are around improving how we use Elasticsearch to obtain higher quality results. 
+
+#### Combination of queries
+The main insight is that no unary query can sufficiently express all the ways a query text matches a document.
+
+These include (at least):
+- exact matches
+- exact matches with spelling mistakes (e.g. "ordered_iems" instead of "ordered_items")
+- analyzed matches where a user inputs keywords (e.g. "returns") and looks to find the most relevant entities
+
+Elasticsearch uses [compound queries](https://www.elastic.co/guide/en/elasticsearch/reference/current/compound-queries.html) to combine multiple subqueries together. The most common is  [Boolean](https://www.elastic.co/guide/en/elasticsearch/reference/master/query-dsl-bool-query.html).
+
+#### Query Optimizations
+To be effective, the union of queries discussed above still needs to be composed of high quality subqueries. 
+Here, we’ll list a handful of optimization directions and parameters we’ll experiment with:
+
+##### Fuzziness
+Allows for misspellings by specifying a maximum edit distance. See [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-fuzzy-query.html).
+
+##### Synonyms 
+This will require tuning and validation, but some users may get significant mileage from a hand-generated synonyms dictionary file. 
+We could provide a facility for specifying such a file. [Docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-graph-tokenfilter.html).
+
+#### Match Phrase Query
+In case someone searches for a phrase out of an entity description, this subquery will take word order into consideration. [Docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query-phrase.html).
+
+#### Index Fields in More Ways
+Experimenting with multiple analyzers (i.e. independently analyzing an input field via multiple analyzers) has potential benefits like normalizing stemmings and filtering stop words. 
+See [multi-field](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html#multi-fields) about how to index a field in multiple ways, [english](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html#english-analyzer) analyzer for stemmings and stop word removal.
+Amundsen already does this to index some fields both as an unanalyzed `keyword` and as analyzed `text`.
+
+
+## Drawbacks and Alternatives
+We believe our proposal will be a significant leap in search quality that will broadly benefit all Amundsen users. Along
+with improving out-of-the-box ranking and relevance, we plan to make more parameters tunable by end users.
+
+The main drawback is around the proposed migration to ES 7, which is discussed below.
+
+## Unresolved questions
+A full ES7 migration may be difficult for some users. One option is to keep the existing ES proxy for legacy use, but
+this is complicated by needing to upgrade the Elasticsearch library in the search service to v7+. 
+Since it is recommended to use the same version of the ES lib as the server version, making v6 compatible requests
+with the v7 library may run into unexpected corner cases. 
+
+At the same time, v7 has too many new features to pass up and revamping search against the older v6 version is likely
+suboptimal.
+
+ 


rfcs/000-search-revamp.md
```diff
@@ -0,0 +1,124 @@
...
+
+There is an obvious question of whether multiple values would be AND'ed or OR'ed together. Our goal is to provide
+additional expressiveness without going down the path of supporting full logical expressions. To better understand the problem,
+keep in mind that there's two kind of fields: single-valued fields (table name, schema name) and multi-valued fields (columns, tags, badges).
+Our proposed tradeoff between expressiveness and simplicity is to combine filter values for single-valued 
+fields with OR (e.g. "staging,production" filter applied to the "schema" field)
+whereas for multi-valued fields we'd give users an options to choose AND or OR 
+(a user would choose AND/OR to apply the filters "product_id,user_id" to the "column" field)
+
+#### Negation filtering
```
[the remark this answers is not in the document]
mgorsk1: We are definitely not looking back to using Atlas Search API but there are other Atlas users which might be impacted by removing search proxy. 

Arguments for it would be simplicity of Amundsen implementation and aligned search experience but it might hurt the simplicity of Amundsen and Atlas deployment if some users are satisfied with Atlas search and don't want to go into trouble of setting up ES & running databuilder search sync.

rfcs/000-search-revamp.md
```diff
@@ -0,0 +1,124 @@
...
+    one's problem
+
+Although we believe Amundsen has been particularly lacking in the discovery search category, our proposal is seeking
+to improve the experience of both search types. 
+
+
+## UI/UX Details
+
+Although the bulk of proposed changes will be to search result quality and relevance, we're also proposing
+a handful of UI/UX changes to filters. We're currently working on designs and will share those as soon as they're
```
mgorsk1, line 30: One idea that came up multiple times in our user sessions is better feedback on aggregated search results.

For example if I am looking for `transactions` data and Amundsen returns 1902 results it's still confusing for some users and they'd like to know top 5 schemas from which those results are coming to get `bird's eye view` on the query made by user. It's just one example but introducing aggregates into search revamp could be also part of UX/UI change.

rfcs/000-search-revamp.md
```diff
@@ -0,0 +1,124 @@
...
+There is an obvious question of whether multiple values would be AND'ed or OR'ed together. Our goal is to provide
+additional expressiveness without going down the path of supporting full logical expressions. To better understand the problem,
+keep in mind that there's two kind of fields: single-valued fields (table name, schema name) and multi-valued fields (columns, tags, badges).
+Our proposed tradeoff between expressiveness and simplicity is to combine filter values for single-valued 
+fields with OR (e.g. "staging,production" filter applied to the "schema" field)
+whereas for multi-valued fields we'd give users an options to choose AND or OR 
+(a user would choose AND/OR to apply the filters "product_id,user_id" to the "column" field)
+
+#### Negation filtering
+Another highly requested feature is specifying negative filters - e.g. "schema is not temp, tmp, or temporary".
```
mgorsk1, line 45: Apart from negation, positive boosting is also valid use-case. In ING we hard-coded it on query level (which could be something configurable in search proxy) to promote tables coming from official data sources. So if the table has `ing_official_data` tag it's relevance is boosted in the search results.

This could be something simple like configuration option - list of maps with field, value, boost_factor properties.

rfcs/000-search-revamp.md
```diff
@@ -0,0 +1,124 @@
...
+#### Filtered vs unfiltered search unification
+Filtered search was added on top of regular search and was implemented separately via separate endpoints. 
+It also uses a different (and incorrect) elasticsearch [query_string](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html) command. 
+We plan to unify the implementations, so both use the same top-level elasticsearch command either with or without an additional filter specifier. 
+
+#### Replace JSON in queries/mappings with Elasticsearch library objects
+The requests made to Elasticsearch are constructed with either hardcoded (in the case of mappings) or templated JSON. 
+This makes it more difficult to compose and understand requests, 
+but more importantly it makes it very difficult to extend OSS code. 
+The Elasticsearch-dsl library includes classes for constructing requests by putting together native python objects.
```
mgorsk1, line 74: I think one aspect that could be also raised is how we approach mapping definitions.

Now we have separate mappings for each type of object (table,feature,user,dashboard) and I think the refactor could also improve on that. We can have one mapping (or even better - template) that would span all the fields for all objects. It's a common pattern and it would make reusability of new analyzers/tokenizers/filters much easier.

rfcs/000-search-revamp.md
```diff
@@ -0,0 +1,124 @@
...
+Experimenting with multiple analyzers (i.e. independently analyzing an input field via multiple analyzers) has potential benefits like normalizing stemmings and filtering stop words. 
+See [multi-field](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html#multi-fields) about how to index a field in multiple ways, [english](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html#english-analyzer) analyzer for stemmings and stop word removal.
+Amundsen already does this to index some fields both as an unanalyzed `keyword` and as analyzed `text`.
+
+
+## Drawbacks and Alternatives
+We believe our proposal will be a significant leap in search quality that will broadly benefit all Amundsen users. Along
+with improving out-of-the-box ranking and relevance, we plan to make more parameters tunable by end users.
+
+The main drawback is around the proposed migration to ES 7, which is discussed below.
```
mgorsk1, line 113: I will repeat it here but somehow I don't see search revamp and ES 7 upgrade related as things proposed by this RFC can be achieved by both ES6 and ES7

rfcs/000-search-revamp.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Unresolved questions
+A full ES7 migration may be difficult for some users. One option is to keep the existing ES proxy for legacy use, but
+this is complicated by needing to upgrade the Elasticsearch library in the search service to v7+. 
+Since it is recommended to use the same version of the ES lib as the server version, making v6 compatible requests
+with the v7 library may run into unexpected corner cases. 
+
+At the same time, v7 has too many new features to pass up and revamping search against the older v6 version is likely
```
mgorsk1, line 121: Can you list the features what we would have to pass up to stay ES6 compatible?

rfcs/000-search-revamp.md
```diff
@@ -0,0 +1,124 @@
+- Feature Name: Search Revamp
+- Start Date: 2021-09-20
+- RFC PR: todo
+
+# Search Revamp
+
+## Summary
+
+Revamp search with a cleaner backend implementation, more robust queries, better customizability, and
+making use of modern v7 ES features.
```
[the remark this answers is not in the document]
dkunitsk: I do know it exists but I didn't give it much consideration. My worry was that it'd be too new/different from ES and harder to find resources for, especially documentation which is very good for classic ES. And also I didn't want to accidentally write anything opensearch-specific. But perhaps there's a lot that I'm misunderstanding.

rfcs/000-search-revamp.md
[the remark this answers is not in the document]
dkunitsk: I am a million percent in favor of removing. I'll run some informal polling first, of course, but I think the burden of maintaining it probably far exceeds the combined user burden (if any) of spinning up elasticsearch. 

This is not to mention that the current Atlas proxy supports only a fraction of what's supported by the ES proxy. No user/dashboard/feature search, for starters.

rfcs/000-search-revamp.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Ranking/relevance changes
+The bulk of improvements we want to make are around improving how we use Elasticsearch to obtain higher quality results. 
+
+#### Combination of queries
+The main insight is that no unary query can sufficiently express all the ways a query text matches a document.
+
+These include (at least):
+- exact matches
+- exact matches with spelling mistakes (e.g. "ordered_iems" instead of "ordered_items")
```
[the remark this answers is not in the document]
dkunitsk: That makes sense!

rfcs/000-search-revamp.md
```diff
@@ -0,0 +1,124 @@
...
+it can be hard for users to remember/guess the exact values to filter for. We'd like to provide typeahead
+suggestions that will suggest the most likely filter value given a prefix.
+
+
+## Backend Changes
+As part of this work, most of the elasticsearch [proxy](https://github.com/amundsen-io/amundsen/blob/main/search/search_service/proxy/elasticsearch.py)
+would be re-written, while keeping the API of the search service the same (in part to allow Atlas to stay unchanged).
+
+#### Upgrade Elasticsearch to v7
+Elasticsearch queries/mappings are generally written to a specific version, since a number of features are typically
```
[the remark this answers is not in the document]
dkunitsk: Thank you for pointing this out. You're right I was not too explicit in the RFC, in part to give ourselves more space to experiment. There's two features of v7 in particular that excite us, although we're also motivated by just wanting to move to the latest-and-greatest since we'll be touching so much code anyway.

By far the most exciting feature (which should probably have gotten treatment in the RFC - I left it out because we may want to try this internally before pushing to OSS) is [rank_features](https://www.elastic.co/guide/en/elasticsearch/reference/current/rank-features.html). This [blog](https://www.elastic.co/blog/easier-relevance-tuning-elasticsearch-7-0) gives an excellent overview, and also mentions how `function_score`, which is what we use now, is being replaced by `script_score` for those who need full control (which I don't believe we do). 

What we've hypothesized is that one of our biggest issues with search relevance is lack of good usage metrics that would promote high use tables. Usage is very multifaceted (number of dags, number of daily active users, number of downstreams, number of total queries queries against a table, etc etc) and these usages cannot be simply added together. Right now Amundsen has one `total_usage` field and it does not capture these different usages well.

Also - and this is further out - but in my reading of the `rank_features` docs, it seems that the explicit ability to index a sparse metrics opens the door to an MVP of personalized search. If we can index per-user usage into each document, then personalized search becomes a matter of 1 extra line in the query to take `usages.dkunitskiy` into account, alongside `usages.num_downstreams`, `usages.daily_active_users` etc. 

The other v7 addition is the [revamped](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-as-you-type.html) search-as-you-type suggestions, which are significantly different in v7. I'm not an expert on this, but we plan to follow the docs to make the filterable fields `search_as_you_type` fields to get filter completions. 

In general you're right that v6 is not deprecated and in my understanding, staying on older ES versions is quite normal. And we also are hesitant to make anyone undergo an ES migration. So we are looking for input on the best approach here, ideally balancing short and long term pain.

rfcs/000-search-revamp.md
dkunitsk: Yes! We do something very very similar at Lyft. I also left this out of the RFC (since it's more of a user-configurable feature). Our plan, once queries are composed of extendable native python objects and not templated JSON, is to use [Rescoring](https://www.elastic.co/guide/en/elasticsearch/reference/7.2/search-request-rescore.html) for this. I.e. you'd provide a `rescore` query to match docs that have the `ing_official_data` tag and then provide coefficients of how to combine the score of the `rescore` query vs the original query. 

FYI rescoring isn't natively supported by the python lib but it can be injected as an `extra`, see [issue](https://github.com/elastic/elasticsearch-dsl-py/issues/311).

rfcs/000-search-revamp.md
dkunitsk: This is an excellent suggestion and my instinct is strongly in favor. Let me think on this and draw up the pros/cons and how to best implement.

rfcs/000-search-revamp.md
dkunitsk: Closing in favor of the comment above, so discussion on this question is in one place.

rfcs/000-search-revamp.md
dkunitsk: Closing in favor of the comment above, so discussion on this question is in one place.

rfcs/000-search-revamp.md
dkunitsk: Very solid direction. Would be a huge improvement, and we hear the same feedback around our parts. At the moment I'm not sure we can fit this into the scope of our proposed enhancements, but let me run this by a few folks. At the very least, we can plan ahead and think about how to add this down the road so we don't make suboptimal choices now.

rfcs/000-search-revamp.md
mgorsk1: hm the idea with rescoring is fine idea but limiting in a sense that it doesn't span `all` of the results but only top N (10 by default) from each shard (afair there is only 1 shard by default in our mappings and our use case will very rarely require more than that). so `negating` or `promoting` would only touch top 10 of results from the original es scoring.

dkunitsk: I had figured this was not much of a limitation as one could adjust N to a point at which one can say "hey, if the result wasn't in the top 500 to begin with, maybe it shouldn't get boosted at all". But we're not married to this implementation - we're looking to let users tweak scoring according to their custom logic in a clean and extendable way (that plays well with the original document score instead of just overriding it) and we're open to suggestions.

rfcs/000-search-revamp.md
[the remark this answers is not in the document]
mgorsk1: I think we are fine removing it as long as we provide some transition period so users have comfortable time for switching into new approach. We'd promote the approach that is already outlined in docs (using databuilder search extractor).

dkunitsk:
@mgorsk1 @feng-tao putting this in final comment period. We're getting started on the implementation. I want to settle on the ES v7 question. One thought is we could avoid this for now by keeping the core functionality both v6 and v7 compatible and putting the more interesting features I [outlined](https://github.com/amundsen-io/rfcs/pull/43#discussion_r714148737) behind some config flags. 

I don't love it though - all things being equal, I'd rather make a clean break. But that's not without pain so I'm open to a hybrid approach.

feng-tao:
actually are we talking about single  ES index for all types? I didn't have time to read through all discussions, but if this is the case, we discuss more in details.

mgorsk1:
the próbował was single index **template** and keep multiple indices (all based on the same template)

dkunitsk:
@feng-tao @mgorsk1 Hey, we did think more on this. Quite agree on the problem @mgorsk1 pointed out - we specify the common fields N times, with random differences here and there. 

I am not 100% convinced that introducing templates is worth it - seems like a bit more overhead and more place for error. Instead, we were thinking we'd make some base entity object class, which will be possible once we migrate the hardcoded JSON into Py library object. 

But, confession: I have not used templates before so I'm not an expert on that. And my first time googling Elastic Common Schema was after seeing it in the comment above 😄

mgorsk1:
I understand how using python classes might also be used with this revamp but that really doesn't change the approach just moves from using json mappings to using python classes which in turn is something I feel more python-native than elastic-native. It feels like ORM so is attractive from developers perspective but not quite the same thing and it's definitely more cumbersome to fine-tune es mapping than using raw mapping definition. 

I am not really sure what it means that introducing templates introduces room for more error. For me it's the opposite - having one place enforces unification and singular approach for common fields, mapping settings etc. instead of repeating it multiple times. It's also something Elastic ecosystem promotes for years now with templates being at the heart of every subproject (metricbeat, filebeat, etc.).

dkunitsk:
@mgorsk1 I think my concern is something like this: we'll still need some definition (in code) of the common fields and their types, right? Call this M. And, in each databuilder run, we can't assume the right template(s) already exist in Elasticsearch. 

So then, in a databuilder run, instead of creating a new index based off M, we first create a new template based off M and then create a new index using that template. So I'm not sure I understand where the benefit is to be found. 

Perhaps I am misunderstanding a few things and just need a bit more nudging.

PR #44, opened 2021-09-28
Extend badges browse capabilities

mgorsk1:


rfcs/044-browse-badges.md
```diff
@@ -0,0 +1,61 @@
...
+- Leave current state as-is with users not being able to see all badges defined in Amundsen quickly.
+- Alternative UI approach would be to rename `Popular tags` section to `Popular labels` with `tags` and `badges` subsections
+
+## Prior art
+
+- This topic has been brought up by users several times on our Slack
+
+## Unresolved questions
+
+- Do we show `Popular badges` or just `Badges` (all badges) on landing page (number of unique badges might be significantly
```
dechoma, line 54: +1 to popular badges

rfcs/044-browse-badges.md
```diff
@@ -0,0 +1,61 @@
...
+smaller than tags so maybe there is no point in showing just `popular` ones)?
+- Would it make more sense to have `Popular badges` on landing page above or below tags?
+- Shall we hide `Popular/All badges` section on landing page when `/badges` endpoint returns empty list?
```
dechoma, line 57: I would hide it, some may not use the badge at all

dechoma (commented):
I like this idea, I think it will improve badge usage user expierience

rfcs/044-browse-badges.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Summary
+
+This RFC proposes to introduce `Popular/All badges` section below `Popular tags` on landing page and & browse section.*
+
+! `Popular/All` is either this or that choice that has been explained in `Unresolved questions` section.
+
+## Motivation
+
+After introducing refactor of tags and badges only the former ones get dedicated section on landing page and . As badges
```
dkunitsk, line 16: could you link to what you're mentioning. I'm out of the loop.

mgorsk1: This referred to some Slack threads that were removed due to retention period. Which part you're interested in?

NiravLangaliya:
Can we extend use of badges as Business glossary / Global Taxonomy by organizing them at different category?

NiravLangaliya:
yes, display a badge category would be very good addition. Also We can refer and take some of inspiration from here where they have defined terms to represent as business glossary and also we can see all entities attached to that terms along with hierarchy of it. 

Also they have sperate out information like primary key or partition column as separate information so terms can independently work. but We can expand modeling of Badges to accommodate similar information. 

https://demo.datahubproject.io/glossary/urn:li:glossaryTerm:ClientsAndAccounts.Account/related%20entities 
https://demo.datahubproject.io/dataset/urn:li:dataset:(urn:li:dataPlatform:mysql,User.UserAccount,PROD)/Schema?is_lineage_mode=false

https://github.com/linkedin/datahub/blob/master/docs/rfc/active/1842-business_glossary/README.md

rfcs/044-browse-badges.md
mgorsk1: there is no 'count' attribute yet so for the first stage I'd go with `Available` badges

PR #45, opened 2022-01-25
Adds RFC for tour component

Golodhros:


rfcs/045-amundsen-tour-and-feature-walkthrough.md
```diff
@@ -0,0 +1,94 @@
...
+This modal window will have a "Dimiss" button that would hide the Tour altogether; a "Back" button that would move the user to the previous tour step, a "Next" button that moves it forward and a "Close" button with the usual "X" shape in the top right corner. You can see here [an example](https://react-joyride.com/) of what this tour can do (click on the "Start" button).
+
+For Amundsen maintainers, we will extend the JavaScript configuration file with a block about the tour. This object will have a shape like this:
+
+```js
+...,
+tour: [
+    {
+        path: '/',
+        isFeatureTour: false,
```
allisonsuarez, line 31: what does `isFeatureTour` do when its true? Also if `isShownOnFirstVisit` is false does it show up every time?

Golodhros: `isFeatureTour` would help us differentiate the case where we add a tour for only one feature VS add a tour for the whole page. This will change some of the general props of the component.

If `isShownOnFirstVisit` is false, it won't show unless we click a button to explicitly turn on the tour!

rfcs/045-amundsen-tour-and-feature-walkthrough.md
```diff
@@ -0,0 +1,94 @@
...
+        isShownOnFirstVisit: true,
```
mgorsk1, line 32: how do we decide user has already seen it? is it via a cookie or persisting this in a db?

rfcs/045-amundsen-tour-and-feature-walkthrough.md
```diff
@@ -0,0 +1,94 @@
...
+        isShownProgramatically: true,
```
mgorsk1, line 33: could you explain what it does?

mgorsk1 (changes requested):
I cannot imagine anyone would be against it. Thanks for doing it!

mgorsk1 (approved):

rfcs/045-amundsen-tour-and-feature-walkthrough.md
Golodhros: We'll keep a key/value entry in the user's browser localStorage

rfcs/045-amundsen-tour-and-feature-walkthrough.md
Golodhros: That would be to include a button to show the tour at any time. This will be good for the overall tours for a page, but not so interesting for feature specific features.

PR #46, opened 2022-03-25
Column details access and side panels

knowlb:
[CLICK](https://github.com/amundsen-io/rfcs/blob/f36f710f8a1a055616e7789764d538687d56f973/rfcs/046-column-details-access-and-side-panels.md) to read formatted rfc.


mgorsk1:
Did you also consider merging concepts of Machine Learning Feature and Column and provide a dedicated view for column? For me in many ways ML Feature is very similar (if not identical) to Column - If I had to figure out how would dedicated column view look like, it would be 99% in line with what ML Feature view provides.

rfcs/046-column-details-access-and-side-panels.md
```diff
@@ -0,0 +1,120 @@
...
+The size of the table showing the column info will be reduced when the right panel opens to ensure the column types and actions are not hidden behind it. When the side panel opens, the table area will
+* Column names, nested column hierarchy (e.g. indentation), description, column type and action button will stay in the table area
+* Usage and badges will be hidden from the table area, but displayed in the right side panel. 
+
+
+## Drawbacks
+
+We have held off on major layout changes due to implementation cost and disruption of existing UI, but the column list area is getting crowded as we add more column metadata, and is not flexible enough to add the [nested columns improvements](./047-nested-columns.md) we want to make.
+
+## Alternatives
```
mgorsk1, line 66: I think what is also worth consideration is merging concepts of Machine Learning Feature and Column treat them as the same first-class citizen (with ML Feature as column type the same way we have same view for table and table view). For me in many ways ML Feature is very similar (if not identical) to Column - If I had to figure out how would dedicated column view look like, it would be 99% in line with what ML Feature view provides.

It also opens up possibility to take this even further in the future with providing actual resource hierarchy, something we found to be a limitation in Amundsen (like more clearer database > table > column hierarchy in the app with dedicated views) but would be helpful.

I like the idea of quick access to column details but from the perspective of giving it more space in the UI. Side pane, however, is not very scalable in my opinion - for example: if we wanted to provide graphs for column quality (like distribution of values etc.) it won't be very easy to do with sections that narrow.

Golodhros: Re: side panel size.
I think this could be worked around in the future. If we needed more horizontal space, we could move that panel to be "overlapping" the table content and make it so it could grow more.

knowlb: Thanks so much for the feedback- we have considered this but decided the panel would be the optimal direction for now because we don't have enough metadata to show at column level. Even the side panel might be sparse for most columns. 
Also, we think the side panel is a valuable addition to preview content without leaving the page, whether we have a dedicated page for columns or not. The advantage is that you can quickly peek at the metadata without clicking into next level of resource and compare multiple columns in the list. As you see in the Future Possibilities, this can be applied to view table metadata from search results page. 
What we could do is add a note about possibility to add a dedicated column page, when we feel the need to scale due to overflowing column details in side panel.

dkunitsk (approved):

Golodhros (approved):
LGTM

PR #47, opened 2022-03-25
Nested columns

knowlb:
[CLICK](https://github.com/amundsen-io/rfcs/blob/9c3fdbb00fde83445e517a5509f5347ec83b8792/rfcs/047-nested-columns.md) to read formatted rfc.


rfcs/047-nested-columns.md
```diff
@@ -0,0 +1,189 @@
...
+The hierarchy of nested objects without named fields (e.g. Array and Map) are now better specified and distinguished from nested objects with named fields (e.g. Struct).
+
+
+## UI/UX-level Explanation
+
+### Expand/collapse all + each level drill down
+
+![Default all expanded view](../assets/047/default-all-expanded-view.png)
+
+* As default, all nested columns will be expanded, showing the option for user to ‘Collapse all nested’
```
mgorsk1, line 48: this could be a config option, some users would probably prefer collapsed by default

rfcs/047-nested-columns.md
```diff
@@ -0,0 +1,189 @@
...
+* For most complex tables, this default behavior should still load fine, but in the edge case that there are prohibitively many columns to render (threshold will likely be ~1046), we may need to start with an all collapsed view for performance reasons.
```
mgorsk1, line 49: - does it mean that collecting nested column info will be a separate api call to metadata backend?
- how did you come up with this 1046 number?

rfcs/047-nested-columns.md
```diff
@@ -0,0 +1,189 @@
...
+
+![Each level drill down using chevron](../assets/047/each-level-drill-down-using-chevron.png)
+
+* Users can keep clicking on the chevron to expand each level of nesting if they want to  drill down the path without expanding all nesting at once.
+
+
+
+![Column details in right side panel](../assets/047/column-details-in-right-side-panel.png)
+
+* Clicking on a column name opens the right side panel to display all column details, including badges, description, and stats.
```
mgorsk1, line 70: should we mix changes from different rfcs? this is based on assumption that #46 is integrated in it's current form

rfcs/047-nested-columns.md
```diff
@@ -0,0 +1,189 @@
...
+
+This project comes from the user feedback and shortcomings of our current POC for supporting nested columns.
+
+## Unresolved questions
+
+Decide whether we need to load a collapsed view as default in the edge case that there are prohibitively many columns to render, and what the threshold should be.
+
+## Future possibilities
+
+We could add a breadcrumb at the top of right side to help users see the whole path with clickable shortcuts:
```
mgorsk1, line 185: I really like this and would love to see resource/hierarchy with breadcrumbs more incorporated into Amundsen (already mentioned in #46 with database > table > column relationships)

rfcs/047-nested-columns.md
```diff
@@ -0,0 +1,189 @@
...
+![Column details in right side panel](../assets/047/column-details-in-right-side-panel.png)
+
+* Clicking on a column name opens the right side panel to display all column details, including badges, description, and stats.
+* Only the columns with metadata will be displayed in interactive (Indigo 60) color, so that users can tell whether they will see useful information by clicking on its name.
+* For the new right side panel and layout, see details in the [Column details access and side panels RFC](./046-column-details-access-and-side-panels.md).
+* In the edge case so many nested levels indent the list too much, users can still side scroll or widen the browser. 
+
+
+
+### Nested object without named fields (e.g. Array and Map)
```
mgorsk1, line 77: I am not 100% sure why (and whether) we need to distinguish it at all? This might overcomplicate the UI without adding much value.

rfcs/047-nested-columns.md
Golodhros: Makes sense! That could be easily added.

rfcs/047-nested-columns.md
Golodhros: We weren't thinking about changing the API calls.
That number is a bit arbitrary, but it might be the one we are using right now. We'll test it with the worse scenario we have at Lyft and adjust.

rfcs/047-nested-columns.md
Golodhros: Yes, this is dependent on #46, we don't think we can do this without the other.

rfcs/047-nested-columns.md
kristenarmes: To expand on the response about the API, the nested column information will be stored as a part of the ColumnMetadata, so it will be available from the existing call to the getTableData API

rfcs/047-nested-columns.md
kristenarmes: If a column is a struct, and another column is an array of structs or even an array of arrays of structs (and so on), in the current state the UI would represent them the same way by only showing rows in the table for the named values in the struct. The thought here was by adding a short row for each level it wouldn't need to take up much screen space and you'd be able to quickly distinguish the different column types. In addition, this would allow the possibility of adding and viewing metadata such as badges and descriptions for the intermediate levels.

rfcs/047-nested-columns.md
dkunitsk: Although this one is depending on #46, the reverse is not true. When we were drafting this, it started as one RFC but the content was clearly along two axes and we thought it would be more clear to separate.

rfcs/047-nested-columns.md
dkunitsk: It sounds like a distinction without a difference, but in the half-dozen design meetings we had to discuss how to represent nested structures we found this distinction was quite fundamental.

rfcs/047-nested-columns.md
knowlb: Oops, sorry about the typo on 1046. I meant to say ~1000. We are estimating that even the most complex tables won't have more than 1000 nested columns. This is just to say except for edge cases, we want to show all expanded view as default, but may need special handling for huge tables in case they don't load correctly.

dkunitsk (approved):

Golodhros (approved):
LGTM

